PR #13, opened 2016-02-19
Simplify user full name method

dfaulken:
I may have more substantive suggestions later.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
   scope :active, -> { where(active: true) }
 
   def full_name
-    return "#{self.first_name} #{self.last_name}"
+    [first_name, last_name].join ' '
   end
 
   # Can take 1, 2, or 3 params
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
   scope :active, -> { where(active: true) }
 
   def full_name
-    return "#{self.first_name} #{self.last_name}"
+    [first_name, last_name].join ' '
```
jgthibodeau: Is this purely to change it from a " string to a ' string?

dfaulken: There's no difference between the two as return values - I used single quotes here because double quotes weren't needed. There's two Ruby style guidelines that stood out to me with the old method:

+ Unneeded `return`
+ Unneeded use of `self`.

Now, personally, I like array joining better because it requires the use of fewer symbols, but string interpolation is perfectly fine too (it's what I wrote in the test to avoid logic coupling) - and string interpolation does require double quotes.

jgthibodeau: Ah, makes sense

PR #15, opened 2016-02-19
Incorporate Rubocop

dfaulken:
Most of the failures that are remaining have to do with permissions.

I agree with Rubocop that these methods are pretty complicated.

If you want to incorporate Rubocop without having those problems cause failing builds, you can do:

```
rubocop --auto-gen-config
```

which will generate a `.rubocop_todo.yml` file. If you `inherit_from` this file in `.rubocop.yml`,
rubocop will pass, and as you resolve each of the individual problems, you can remove the lines
which cause that cop not to be run, etc.


dfaulken:
Forgot to mention - all of the changes from 32f8763 were automatically made by running

```
rubocop -a
```

which is short for 'autocorrect'.

PR #22, opened 2016-02-22
User, Group, Permission model specs

jgthibodeau:


spec/factories/group.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :group do |f|
+    f.name 'group'
+    f.description 'this is a group'
```
dfaulken: I see from your group test that names should be unique - should `name` be a sequence then? 

Also, you can define factories without the block variable, e.g.

```ruby
FactoryGirl.define do
  factory :group do
    sequence(:name){ |n| "Group #{n}" }
    description 'Description'
  end
end
```

spec/models/user_spec.rb
```diff
@@ -19,10 +19,80 @@
...
+    # params[:controller], params[:action], params[:id]
+    it 'returns true if the user has a permission with the requested controller, action, and no id_field' do
+      user = create(:user)
+      group = create(:group)
+      permission = create(:permission, controller: 'user', action: 'show', id_field: nil)
+
+      group.permissions << permission
+      user.groups << group
+
+      expect(user.has_permission?(permission.controller, permission.action, nil)).to eq true
```
dfaulken: For this and other predicate methods that start with `has_`, you should be able to use RSpec's dynamic predicate method matchers, e.g.:

```ruby
 expect(user).to have_permission permission.controller,
                                 permission.action,
                                 nil
```

dfaulken: (Shameless plug: if you're interested in incorporating a RuboCop style check to make sure you're using dynamic predicate method matchers, [there's a gem for that](https://github.com/umts/custom-cops)!)

jgthibodeau: That's pretty snazzy

PR #24, opened 2016-02-26
finished fee_schedule management page, and added a model item_type management page

xliu3:
screenshot 2016-02-26 09 17 26

screenshot 2016-02-26 09 17 43


jgthibodeau:
Looks like the spec errors are being caused by link_to not having a current_user in request specs. Currently working towards a solution.

Also, take a look at the codeclimate report, most of it is just indenting/spacing or similar code (the similar code stuff can be ignored for now). If you want to generate this report for yourself you can run `rubocop` which will highlight general styling things.

db/migrate/20160222142237_create_item_types.rb
```diff
@@ -0,0 +1,10 @@
+class CreateItemTypes < ActiveRecord::Migration
+  def change
+    create_table :item_types do |t|
+      t.string :name
+      t.references :fee_schedule
```
phantompatrol: I would also add an index here!

db/schema.rb
```diff
@@ -11,7 +11,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160218175640) do
+<<<<<<< HEAD
+
+=======
+>>>>>>> 7320397a0b2224e2514f2df1001da779f573abee
```
phantompatrol: :checkered_flag:

app/models/fee_schedule.rb
```diff
@@ -0,0 +1,4 @@
+class FeeSchedule < ActiveRecord::Base
+  #need to one to one relationship with car type table.
+  belongs_to :item_type
+end
```
phantompatrol: Recommendation: Please add validations for all `fee_schedule` attributes

jgthibodeau:
Solution to request specs failing: delete autogenerated request specs. They are only really testing routes, but as @phantompatrol pointed out, this is implicitly done by controller specs and we do not have outside services pointing back at us.

app/models/item_type.rb
```diff
@@ -0,0 +1,4 @@
+class ItemType < ActiveRecord::Base
+  has_one :fee_schedule, dependent: :destroy
+
+end
```
phantompatrol: Ditto with validations for on attributes

dfaulken:
To what extent is item type information going to be pulled from `aggressive-epsilon`? @phantompatrol @jgthibodeau

spec/controllers/fee_schedules_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+#
+# It only uses APIs available in rails and/or rspec-rails.  There are a number
+# of tools you can use to make these specs even more expressive, but we're
+# sticking to rails and rspec-rails APIs to keep things simple and stable.
+#
+# Compared to earlier versions of this generator, there is very limited use of
+# stubs and message expectations in this spec.  Stubs are only used when there
+# is no simpler way to get a handle on the object needed for the example.
+# Message expectations are only used when there is no simpler way to specify
+# that an instance is receiving a specific message.
```
jgthibodeau: Please delete unneeded comments and fill in missing specs

xliu3: okay!

jgthibodeau:
@dfaulken We are currently planning to pull the names/ids of the available item types so that we can link fee schedules to specific types for billing on our end

dfaulken:
@jgthibodeau So then do you still have a need for an ItemType model on your end?

jgthibodeau:
@dfaulken I guess not, we could roll the information held there into the fee_schedule table itself. fee_schedules will link to an item_type_id from the api, so when a reservation for a specific item_type is made we can apply the correct fee on our end. We would just need to ensure that fee_schedules have a unique item_type_id

phantompatrol:
Having the scaffolding on our end makes things a bit easier for now, but regarding https://github.com/umts/probable-engine/pull/24#issuecomment-190430082 specifically, that might be our first 'in' with these two applications. We are basing a lot of our billing and availability on item type, so this might be as good of an opportunity as ever to officially link these apps.

Storing a [UUID](http://ruby-doc.org/stdlib-2.2.1/libdoc/securerandom/rdoc/SecureRandom.html#method-c-uuid) from the API should be sufficient to preserve referential integrity.

@dfaulken @jgthibodeau, have you considered how development and testing environments will interact with the API, at least in the short term?

dfaulken:
For tests, I'd suggest mocking based on [the documentation](https://github.com/umts/aggressive-epsilon/blob/master/README.md).

For development, perhaps spin up an instance of aggressive-epsilon locally and point your probable-engine instance to that? Gives us an opportunity to test API keys, etc.

@phantompatrol, would you recommend on the aggressive-epsilon side that we go with something like a UUID rather than just a primary key for referencing reservations, items, item types, etc.?

phantompatrol:
@dfaulken UUIDs, I would. Though I would not go so far as to suggest using them as a primary key.

dfaulken:
Will do, opened https://github.com/umts/aggressive-epsilon/issues/17.

Issue #9, opened 2016-02-12
Rental Status changes


jgthibodeau:
Rental status changes need to switch around as things update for a rental

PR #25, opened 2016-02-26
Rental States

josephkelley:
Basic state machine with [aasm](https://github.com/aasm/aasm) for #9.


app/models/rental.rb
```diff
@@ -0,0 +1,38 @@
...
+    state :canceled
+
+    event :cancel do
+      transitions from: :reserved, to: :canceled
+    end
+
+    event :pickup do
+      transitions from: :reserved, to: :checked_out
+      after do
+        self.checked_out_at = Time.zone.now
```
phantompatrol: is `self` required?

dfaulken:
Hey @josephkelley! This is Dave over at Transit - since the combination of this app and [aggressive-epsilon](https://github.com/umts/aggressive-epsilon) is a joint programming venture between Parking and Transit, @jgthibodeau has graciously allowed me to take a look at the pull requests to this app, specifically the specs.

db/migrate/20160217192112_create_rentals.rb
```diff
@@ -0,0 +1,14 @@
+class CreateRentals < ActiveRecord::Migration
+  def change
+    create_table :rentals do |t|
+      t.string :rental_status
```
phantompatrol: please index me

spec/models/rental_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe Rental, type: :model do
+  describe 'rental_status' do
+    before(:each) do
+      @rental = Rental.create(user_id: 0, department_id: 0, reservation_id: 0, fee_schedule_id: 0)
```
phantompatrol: Factory girl?

spec/models/rental_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe Rental, type: :model do
```
dfaulken: Since RSpec is configured with `infer_file_type_from_spec_location!` ([source](https://github.com/josephkelley/probable-engine/commit/e58b45063c538816c0378bedc996ec914c89d38a#diff-e11bdaf9116be3911e9557b69e242370R15)), you can actually just write

```ruby
describe Rental do
```

here.

spec/models/rental_spec.rb
dfaulken: :+1:

dfaulken: Once that's set up, I would also recommend a `let` block here so that you don't need to deal with instance variables, e.g.:

```ruby
let(:rental) { create :rental }
```

dfaulken:
General spec comment: since states are a pretty simple thing to look at, you can trim down the duplication in the specs a little like this: https://gist.github.com/dfaulken/f30d67e647eaf2a432e5.

app/models/rental.rb
josephkelley: this should be an update also

db/migrate/20160217192112_create_rentals.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.integer :user_id, null: false
+      t.integer :department_id, null: false
+      t.integer :reservation_id, null: false
+      t.integer :fee_schedule_id, null: false
+      t.datetime :checked_out_at
+      t.datetime :checked_in_at
```
jgthibodeau: Are `checked_out_at` and `checked_in_at` still needed with the statemachine setup?

spec/models/rental_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it 'is inspected after inspect' do
+      @rental.pickup
+      @rental.return
+      @rental.inspect!
+      expect(@rental).to be_inspected
+    end
+    # or should they just go back to the pool after inspection? maybe not if damaged
+    # should there be a "damaged" state?
```
jgthibodeau: I think this is a question to involve @dfaulken, do we want [aggressive-epsilon](https://github.com/umts/aggressive-epsilon) to track damage and incorporate that information when a reservation is requested from it?

db/migrate/20160217192112_create_rentals.rb
josephkelley: Currently yes. The alternative is to use something like Paper Trail, which wouldn't scale well per @phantompatrol. This is worth investigating further though.

phantompatrol: Just to expound a bit, paper trail is a fantastic auditing tool. But using it to iterate different versions of the object and establish a invoice is comparatively slow / subject to failure in the event of a schema change.

To put it in context, if we use paper trail to establish when the customer checked in/out, we would have to walk the object and fetch the `updates_at` times when the object switches between states. That means we would have to detect the switch, record each time stamp, compare that against the scheduled check times. Doable, but not elegant.

We could always roll this (time stamps related to billing) into its own first class object.

dfaulken: > Doable, but not elegant.

:+1:

PR #29, opened 2016-03-01
Add incidentals

### LINKED ISSUE #3: Incidental management
Incidental management needs to be built up and tested

tomecho:
as far as i can get right now without rentals built, feel free to critique! closes #3


.gitignore
```diff
@@ -17,4 +17,10 @@
 /tmp
 
 # Ignore sensitive config files
-/config/database.yml
\ No newline at end of file
+/config/database.yml
+
+# Ignore editor temp files
```
dfaulken: Since these files are made by your specific editor and system, I'd recommend placing them in a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) in your home directory - there's no guarantee that every developer for this app will need to ignore those files.

app/models/incidental_type.rb
```diff
@@ -0,0 +1,11 @@
+class IncidentalType < ActiveRecord::Base
+  has_many :incurred_incidental
+  validates_presence_of :name
```
dfaulken: Rails 4 prefers the newer validation syntax, e.g.:

```ruby
validates :name, :description, :base, :modifier_amount, :modifier_description, presence: true
```

spec/factories/incidental_type.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :incidental_type do
+    sequence(:name) { |n| "name #{n}" }
```
dfaulken: It's nice for the factory to echo the model validations - so you should use a `sequence` when you have a uniqueness validation, which it doesn't look like is the case here. A static string should be fine - or implement a uniqueness validation.

spec/models/incidental_type_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe IncidentalType, type: :model do
+  context 'will properly do validations' do
+    it 'builds an IncidentalType' do
+      expect(IncidentalType.new(name: 'test', description: 'this isnt a real IncidentalType', base: 2.0, modifier_amount: 2.0, modifier_description: 'two times as much!')).to be_valid
```
dfaulken: It'd be nicer to use FactoryGirl here, e.g.

```ruby
context 'with default attributes' do
  it 'passes validations' do
    expect(create :incidental_type).not_to raise_errors
    #or
    expect(build :incidental_type).to be_valid
  end
end
# Here, we just modify the relevant attribute, and leave the rest in place
context 'with missing attributes' do
  it 'fails validation' do
    expect(build :incidental_type, description: nil).not_to be_valid
    # ...
  end
end
```

spec/factories/incidental_type.rb
tomecho: Actually now that you bring it up a uniqueness validation on a incidental type is probably a good idea, unless we want multiple types of the same name with different fees

jgthibodeau: I'd think we want that to be unique

tomecho: Ok! I changed it.

spec/models/incidental_type_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+    it 'fails to build when it is missing data' do
+      expect(build :incidental_type, name: nil).not_to be_valid
+      expect(build :incidental_type, description: nil).not_to be_valid
+      expect(build :incidental_type, base: nil).not_to be_valid
+      expect(build :incidental_type, modifier_amount: nil).not_to be_valid
+      expect(build :incidental_type, modifier_description: nil).not_to be_valid
+    end
+    it 'wont create two types with the same name' do
+      same = create :incidental_type, name: 'same'
```
dfaulken: It's better to only specify model attributes when there's something significant about the value(s). Since your test should work regardless of what the incidental type's name is, better not to write a specific one.

In particular, I could write the following code:

```ruby
class IncidentalType < ActiveRecord::Base
  validate :name_is_not_same

  # rest of model code

  private

  def name_is_not_same
    if name == 'same'
      errors.add :name, 'cannot be the word same'
    end
  end
end
```

which would make your test pass for the wrong reason.

db/migrate/20160219202933_create_incurred_incidentals.rb
```diff
@@ -0,0 +1,12 @@
+class CreateIncurredIncidentals < ActiveRecord::Migration
+  def change
+    create_table :incurred_incidentals do |t|
+      t.references :incidental_type, index: true, foreign_key: true
+      t.decimal :times_modified
+      t.text :notes
+      t.text :document
```
jgthibodeau: We probably will want to change document to be a separate model instead of a single field, I imagine a check-in officer would want to upload multiple images showing different angles on the same damage

spec/models/incidental_type_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      expect(build :incidental_type, base: nil).not_to be_valid
+      expect(build :incidental_type, modifier_amount: nil).not_to be_valid
+      expect(build :incidental_type, modifier_description: nil).not_to be_valid
+    end
+    it 'wont create two types with the same name' do
+      same = create :incidental_type
+      expect(build :incidental_type, name: same.name).not_to be_valid
+    end
+  end
+end
```
jgthibodeau: May want a sanity check on the fee method in here too, I know it's a simple method but it's a good idea to spec each method

PR #31, opened 2016-03-04
Inventory api mocks

### LINKED ISSUE #30: inventory api mocks
Hey @jgthibodeau, @phantompatrol suggested we implement api fixtures so we are not dependent on aggressive-epsilon for testing.

tomecho:
fixes #30 
@dfaulken might want to look as these mocks to see if they conform to his api

### DIFF
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -1,7 +1,7 @@
 require 'json'
 class Inventory < ActiveRecord::Base
-#this class is all mocked for now
-#in the future it should raise an exception if the api doesnt return a sucess status
+  # this class is all mocked for now
+  # in the future it should raise an exception if the api doesnt return a sucess status
 
   def self.item_types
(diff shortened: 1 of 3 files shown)


tomecho:
lots of code climate issues, should probably correct those

tomecho:
oh woops
i accidently the pull request

tomecho:
adding tests, but for now this should help with building the code that relies on this

dfaulken:
Our README is current, so if this matches that, then it's fine. I'm not sure that an ActiveRecord model is necessarily the best way to go about mocking an API, but that's a decision for your team to make.

tomecho:
@dfaulken being an active record was an accident (force of habit i suppose).  Its just a plain old model.  Is it to early to know the final resting point for the production api?

dfaulken:
At this point it's only going to change if you all need it to.

PR #39, opened 2016-04-04
Rentals can be made

jgthibodeau:


spec/controllers/rentals_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'rails_helper'
+
+describe RentalsController do
+  let!(:rental) { create(:rental) }
+  let!(:rental2) { create(:rental) }
+
+  before(:each) do
+    @current_user = create(:user)
+    controller.instance_variable_set('@current_user', @current_user)
```
dfaulken: If this is going to be common across controller specs, I'd suggest a helper method.

[Here's an example](https://github.com/umts/jobapps/blob/master/spec/spec_helper.rb#L53-L80).

Issue #33, opened 2016-03-08
inventory api


tomecho:
some production code to go along with the mocks

tomecho:
fixed by #48

PR #48, opened 2016-04-12
Add inventory calls

tomecho:


tomecho:
This will not pass travis build because travis does not have an api key .... yet

tomecho:
fixes #33

spec/models/inventory_spec.rb
```diff
@@ -1,107 +1,116 @@
...
+  let(:name) do
+    name = ''
+    name += (rand(26) + 65).chr until name.length > 120
+    name
   end
 
-  it 'can raise an exception' do
-    expect { Inventory.mock_exception }.to raise_error(InventoryError)
+  before :each do
+    Timecop.freeze
```
dfaulken: Out of curiosity, why is this needed?

spec/models/inventory_spec.rb
```diff
@@ -1,107 +1,116 @@
...
+
+    # get all item type
+    expect { response = Inventory.item_types }.not_to raise_error
+    expect(response).to be_a(Array)
+    expect(response.count).not_to eq 0
+
+    # get one item type
+    expect { response = Inventory.item_type(item_type_uuid) }.not_to raise_error
+    expect(response).to be_a(Hash)
+    expect(response.try(:[], 'uuid')).to eq(item_type_uuid)
```
dfaulken: So is `try` needed here? Wouldn't the previous line fail?

```ruby
nil.is_a? Hash # false
```

spec/models/inventory_spec.rb
tomecho: One of the methods in the inventory model searches through reservations
between a time. But I really isnt necessary now that you mention it because
it depends on the remote server as well.

spec/models/inventory_spec.rb
```diff
@@ -1,107 +1,116 @@
 require 'rails_helper'
 include InventoryExceptions
-
-RSpec.describe Inventory, type: :model do
-  before :all do # this is only for the mocks!
-    @uuid = 'not real uuid'
+# these tests must be run in order and all at once
```
dfaulken: Out of curiosity, why is this?

tomecho: Because the test creates the item type, item and then a reservation in
order.

tomecho: At first I wasnt sure this was the best way but after a little bit of
googling people say testing an external api isn't necessary at all. However
I talked with some people in the office and this was the conclusion I came
to.

dfaulken: Cool. Thanks!

PR #50, opened 2016-04-20
Change api config settings

tomecho:
A quick patch to guarantee travis works for future releases. And to make the config a little bit cleaner.


tomecho:
This is super odd, travis likes this build but it fails locally

tomecho:
ok @jgthibodeau pending review this is aok to merge

Issue #44, opened 2016-04-08
Department Management


jgthibodeau:
Need to get department management put together
Billable user is the person who will receive invoices for any rentals made for the department
Not sure if we want this to be an admin only management tool or have it fall into another groups

PR #53, opened 2016-04-27
Departments

josephkelley:
#44


dfaulken:
Probably a good idea to exclude most of `coverage`. [Here](https://github.com/umts/detours/blob/master/.gitignore#L11-L14)'s an example of what that might look like.

spec/controllers/departments_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    context 'with valid attributes' do
+      it 'saves the new department in the databse' do
+        expect do
+          post :create, department: attributes_for(:department)
+        end.to change(Department, :count).by(1)
+      end
+
+      it 'redirects to the department page' do
+        post :create, department: attributes_for(:department)
+        expect(response).to redirect_to Department.last
```
dfaulken: That's a neat trick - I hadn't thought of that.

spec/controllers/departments_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+    end
+  end
+
+  describe 'POST #remove_user' do
+    it 'removes the user from the department' do
+      old_user = create(:user)
+      department.users << old_user
+
+      post :remove_user, id: department, user_id: old_user
```
dfaulken: Better to simulate the data the controller method will actually receive - in this case, the ID of the user (not the whole object).

spec/controllers/departments_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        expect(department.name).to eq(new_name)
+      end
+      it 'redirects to the group page' do
+        new_name = department.name + 'new'
+        post :update, id: department, department: { name: new_name }
+        expect(response).to redirect_to department
+      end
+    end
+
+    context 'with invalid attributes' do
```
dfaulken: Does it return some kind of helpful HTTP status code too? Or maybe an explanation of why the change wasn't made?

dfaulken:
As another general comment, we've had good success using integration tests to achieve a lot of what your controller specs do.

jgthibodeau:
Merging with master should take care of all the coverage files getting pulled back in
Also, can you please check rubocops stuff? Seems like code climate is only yelling about simple formatting things

PR #56, opened 2016-05-16
Fix departments merge

tomecho:


tomecho:
this isnt necessary, I had only broke it locally.

PR #59, opened 2016-05-19
Admin checking process

### LINKED ISSUE #5: Check out/in golfcarts
Show all rentals for the day
Maybe color code green if good to go, red if overdue?

Check out:
- show list of steps to perform
  - document damages (photos)
  - check vehicle charge level
  - confirm return time
  - print pre-filled pdf with rules
  - have both sign it
- submit

Check in:
- show list of steps to perform
  - document damages (photos)
  - sign documents?
  - plug in vehicle
- submit

### LINKED ISSUE #58: rentals should never be destroyed, only deactivated

### LINKED ISSUE #60: digital signature model
will condense all signatures across app

### LINKED ISSUE #61: clean up rental factories
this was totally me, i created a bunch of extras
rental should be valid by default, valid_rental should be unnecessary.
this should remove a bunch of silly things like `create :valid_rental, item_type: create(:item_type, name: 'TEST_ITEM_TYPE')` which is obviously subverts the point of a factory.

There should however be a `create :mock_rental` that does not reach out to the api which will save time and make it easier on the person writing the tests because they wont have to delete any rentals afterwards (they clutter up aggressive-epsilon [alternatively we could just create an absurd amount of items for that particular item type on aggressive-epsilon]).

tomecho:
fixes #5 
fixes #58 
fixes #60  
fixes #61


app/helpers/application_helper.rb
```diff
@@ -79,4 +79,39 @@ def flash_message(type, text, now = nil)
...
+      return 'Reserved future'
+    when 'success'
+      return 'Reserved pickup imminent'
+    when 'warning'
+      return 'Overdue for pickup'
+    when 'danger'
+      return 'Overdue return'
+    when 'info'
+      return 'Ongoing rental'
+    else
```
dfaulken: `else return nil` can also be achieved by just writing none of htat.

app/helpers/application_helper.rb
```diff
@@ -79,4 +79,39 @@ def flash_message(type, text, now = nil)
...
+    elsif rental.rental_status == 'reserved' && rental.start_time.to_date == Time.zone.today
+      return 'success'
+    elsif rental.rental_status == 'reserved' && rental.start_time > Time.current
+      return 'active'
+    else
+      return ''
+    end
+  end
+
+  def rental_status_english(status)
```
dfaulken: Keep in mind that the return value of a method is the value output by the last expression. This is to say, this method could just be:

```ruby
def rental_status_english(status)
  case status
  when 'active' then 'Reserved future'
  # ...
end
```

although, truth be told, this just seems like a mapping from string values to string values, and could be accomplished best perhaps with a Hash.

app/views/check_in_out/_check_in_list.html.erb
```diff
@@ -0,0 +1,27 @@
+<h1>Check In</h1>
```
dfaulken: A little late for this app, perhaps, but we find HAML to be much better except in the most complex of cases.

tomecho:
dont merge this yet
adding more stuff

tomecho:
ok this is good to go, ill expand on digital signature tests in a different branch
(today is my last day so i want to have some solid stuff to get up there)

PR #72, opened 2016-06-17
Rental touchup

owlee:
Adding and modifying interaction and visuals between the rental scheduler and the new rental path. The form for the new rental will be pre-populated whens selecting from the rental schedule.


app/views/rentals/_form.html.erb
```diff
@@ -1,43 +1,50 @@
 <%= form_for @rental, method: :post do |f| %>
   <div class="form-group col-xs-12">
-    <%= f.label :type, 'Item Type', class: 'control-label col-xs-2' %>
+    <%= f.label :type, 'Golf Cart', class: 'control-label col-xs-2' %>
     <div class="col-xs-6">
-      <%= f.select :item_type_id, options_from_collection_for_select(@item_types.order(name: :asc), "id", "name"), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)" %>
+      <%= f.select :item_type_id, options_from_collection_for_select(@item_types.where(name: params["item_type"]).order(name: :asc), "id", "name"), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)", disabled: true %>
```
phantompatrol: I would move the `@item_types.where...` call into the controller on the next iteration.

PR #74, opened 2016-06-22
Update seeds

phantompatrol:
This is a temporary measure as we rethink the connection with the Inventory. Not to be merged.

### DIFF
--- a/db/db_yml/item_types.yml
+++ b/db/db_yml/item_types.yml
@@ -7,8 +7,3 @@
   disclaimer: "Maximum of 6 riders.\nNo joyriding."
   base_fee: "200"
   fee_per_day: "20"
-
-- name: "TEST_ITEM_TYPE"
-  disclaimer: "This is for testing only."
-  base_fee: "200"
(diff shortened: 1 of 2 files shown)


phantompatrol:
@owlee , this is the update to the seeds that we discussed, but I would prefer not to merge this.

LONG story short, I think the inventory model could use some attention and this PR highlights it. Building out the issue here: https://github.com/umts/probable-engine/issues/75

owlee:
screen shot 2016-06-23 at 9 16 26 am

Running seed file produce the error above. I have regenerated the api key and have recreated databases (dropped, created, and migrated) on both Probable and Aggressive. I will be looking more into it.

phantompatrol:
@owlee , checking now

phantompatrol:
Very odd. But for anyone listening in `db:drop/create/migrate/seed` on the local app and api rectified the issue.

phantompatrol:
Closed, in favor of https://github.com/umts/probable-engine/pull/78

PR #78, opened 2016-07-06
Inventory Api Gem

phantompatrol:
- Integration with an inventory API gem
- Item type seeding
- Item seeding

While this is not a final implementation, it is a step much further in the right direction


app/models/rental.rb
```diff
@@ -82,6 +82,7 @@ def delete_reservation
     true
   end
 
+  #TODO: Move logic to a Helper?
```
phantompatrol: https://github.com/umts/probable-engine/issues/79

PR #84, opened 2016-07-08
JQuery Calendar for Admins

owlee:


app/models/rental.rb
```diff
@@ -86,6 +86,36 @@ def times
...
+    case rental_status
+    when 'reserved'
+      return '#0092f'
+    when 'checked_out'
+      return '#f7ff7'
+    when 'checked_in'
+      return '#09ff00'
+    else
+      return '#000000' # black signifies a non event status
+    end
```
phantompatrol: Can this be done with css classes and styles?

owlee: Yes, it certainly should be within the CSS side. However, since I'm working with the gon gem, it will be hacky to break down the gon variable within the rental.js to insert the color. The color is passed in as json, e.g. events: [{title: "1", date: "1-2-16", color: "fff130" }], so no CSS is directly involved. We can discuss if there is a better option?

owlee:
I will be adding JS disclaimer to this request.

app/assets/stylesheets/rentals/rental_schedule.css
```diff
@@ -17,7 +17,8 @@ table, th, td {
 }
 
 .simple-calendar table td {
-  height: 100px !important;
+  height: 123px !important;
+  width : 163px !important;
```
phantompatrol: Does this work without the `!important`? I am curious because this should be pretty high up the priority stack (for CSS).

owlee: It does not because the standard table td is even higher in the stack, the !important 1 ups the default table css.

app/controllers/rentals_controller.rb
```diff
@@ -19,6 +22,7 @@ def show
   # GET /rentals/new
   def new
     @rental = Rental.new
+    @start_date = params['start_date'] ? params['start_date'].to_date : Time.zone.today
```
phantompatrol: For a shorter line, you might be able to do `params['start_date'].try(:to_date) || Time.zone.today`

app/assets/stylesheets/rentals/rental_schedule.css
owlee: Also, may be getting rid of simple calendar in my next pull request in favor of a simpler form for scheduling golf carts.

app/controllers/rentals_controller.rb
owlee: I like it!

app/views/rentals/new.html.erb
```diff
@@ -1,3 +1,7 @@
 <% provide( :page_header, 'New Rental' ) %>
 
-<%= (params["item_type"] ? render('form') : content_tag(:h1, "Please schedule a rental through the Rental Scheduler.")) %>
+<% unless @current_user.has_group? Group.where(name: "admin") %>
```
phantompatrol: Style nit: Please move the AR call to the controller

phantompatrol:
Please remove the `coverage/index.html` & `.DS_Store` files. I have the latter in my global .gitignore

phantompatrol:
The code climate check is right on, btw. JS double equals does type coercion.

app/assets/stylesheets/rentals/rental_schedule.css
phantompatrol: Fair enough. WIP

app/views/rentals/_form.html.erb
```diff
@@ -2,28 +2,21 @@
   <div class="form-group col-xs-12">
     <%= f.label :type, 'Golf Cart', class: 'control-label col-xs-2' %>
     <div class="col-xs-6">
-      <%= f.select :item_type_id, options_from_collection_for_select(@item_types.where(name: params["item_type"]).order(name: :asc), "id", "name"), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)", readonly:true %>
+      <%= f.select :item_type_id, options_from_collection_for_select(@item_types, "id", "name"), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)" %>
```
phantompatrol: From @owlee: Check `update_disclaimer`

Issue #79, opened 2016-07-06
Move date formatting into a helper


phantompatrol:
This line in the renal model is used exclusively in view logic, please move it to a helper.

https://github.com/umts/probable-engine/blob/master/app/models/rental.rb#L86

PR #87, opened 2016-07-13
moving basic_info to helper

cmednieks:
regarding #79


phantompatrol:
Apologies to @cmednieks , upon reflection in code review we think the previous implementation was OK.

Issue #91, opened 2016-07-14
As a CSR, I do not need to sign the processing form


phantompatrol:
As a CSR, I do not need to digitally sign the processing form. I have already logged in and that should be credential enough.

phantompatrol:
Completed by @cmednieks

PR #96, opened 2016-07-14
Removing csr signature from check-in/check-out process

cmednieks:
see issue #91


phantompatrol:
@cmednieks, would you please push a new coverage/.last file up, with the passing specs? The coverage did drop a bit, but that is just because you removed lines (and thus there was less code to test)

phantompatrol:
Other than that, this is 🚢

PR #102, opened 2016-07-18
Financial Transaction summary added to rental

xliu3:


app/models/financial_transaction.rb
```diff
@@ -3,4 +3,8 @@ class FinancialTransaction < ActiveRecord::Base
   belongs_to :transactable, polymorphic: true
 
   validates :rental_id, presence: true
+
+  def self.get_rental_summation(rental_id)
+    FinancialTransaction.where(rental_id: rental_id).sum(:amount)
+  end
```
dfaulken: Might this be more naturally implemented as a `Rental` model method?

```ruby
def summation
  financial_transactions.sum :amount
end
```

app/views/rentals/transaction_detail.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <th>Transaction Adjustment</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @financial_transactions.each do |ft|%>
+        <tr>
+          <td><%= ft.created_at %></td>
+          <td><%= ft.transactable_type %></td>
+          <td><%= number_to_currency ft.amount if ft.amount != nil %>
+              <%= number_to_currency 0 if ft.amount == nil %></td>
```
phantompatrol: How do folks feel about using `number_to_currency (ft.amount || 0)`?

spec/models/financial_transaction_spec.rb
```diff
@@ -57,4 +57,23 @@
       expect(incidental_trans.amount).to eq(incidental.incidental_type.base+(incidental.times_modified*incidental.incidental_type.modifier_amount))
     end
   end
+
+  describe '#self.get_rental_summation' do
+    before(:each) do
+      @item_type = create :item_type, name: 'TEST_ITEM_TYPE'
+      @rental = create :rental, item_type: @item_type
+      @incidental = create :incurred_incidental, rental: @rental
+    end
```
phantompatrol: The [rental factory](https://github.com/umts/probable-engine/blob/master/spec/factories/rental.rb#L5) has an [association](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#associations) that creates an `ItemType` automagically.

Further, since you are not using `@incidental`, you can create the rental in line

Becomes something to the effect of:
```Ruby
before(:each) do
  @rental = create :rental
  create :incurred_incidental, rental: @rental
end
```

If you were feeling as though you wanted to prune out an amazing amount of code, refactoring the [incurred_incidental factory](https://github.com/umts/probable-engine/blob/master/spec/factories/incurred_incidental.rb) to eliminate the mock object could result in this section of code becoming:
```Ruby
before(:each) do
  @rental = create(:incurred_incidental).rental
end
```

spec/models/financial_transaction_spec.rb
```diff
@@ -57,4 +57,23 @@
...
+
+    it 'return the sum of financial transactions which have the same rental' do
+      result_from_function = FinancialTransaction.get_rental_summation(@rental.id)
+      sum_ft_by_rental = 0
+
+      FinancialTransaction.where(rental: @rental).each do |ft|
+        sum_ft_by_rental += ft.amount
+      end
```
phantompatrol: Consider using [`inject`](http://apidock.com/ruby/Enumerable/inject) to turn it into something like:
`@rental.financial_transactions.inject(0) do |memo, ft|...`


Sorry for sending you to apidock, ruby doc is down:
screen shot 2016-07-19 at 11 33 20 am

app/controllers/incurred_incidentals_controller.rb
```diff
@@ -12,6 +12,8 @@ def new
   end
 
   def edit
+    params[:rental_id] = @incurred_incidental.rental.id
```
dfaulken: A bit odd to be manually editing request parameters to then pass to the view - seems like a function which would be served more naturally by instance variables.

spec/models/financial_transaction_spec.rb
dfaulken: Or perhaps

```ruby
sum_ft_by_rental = @rental.financial_transactions.map(&:amount).inject(:+)
```

phantompatrol: Side note:

```
irb(main):014:0> f = FinancialTransaction.create(rental: Rental.last)
...
=> #<FinancialTransaction id: 5, rental_id: 3, transactable_id: nil, transactable_type: nil, amount: nil, adjustment: nil, note_field: nil, created_at: "2016-07-19 15:53:23", updated_at: "2016-07-19 15:53:23">
irb(main):015:0> f.amount
=> nil
```
Added to https://github.com/umts/probable-engine/issues/103

Issue #92, opened 2016-07-14
As an admin, I should not see upcoming rentals that were not picked up


phantompatrol:
(See Title) These rentals were essentially never picked up. Time of photo taken 7/14, note how unprocessed rentals still appear.

![screen shot 2016-07-14 at 10 33 58 am](https://cloud.githubusercontent.com/assets/129601/16843279/9373970e-49ae-11e6-9362-3f1d8b55ffe5.png)

Please move those rentals to `Past Rentals`

cmednieks:
I will work on this

PR #110, opened 2016-07-20
Changing home view

cmednieks:
- Added separate tables for ongoing rentals and no-show rentals(rental status is still reserved but the end date has passed)
- Added separate tab views for each table
- Users in admin group can see all rentals, while others can only see the rentals they've made
- Makes changes referencing issue #92 

screen shot 2016-07-20 at 10 11 49 am


app/controllers/home_controller.rb
```diff
@@ -4,8 +4,25 @@ class HomeController < ApplicationController
...
+    categorize_rentals
+    non_admin_filter unless @current_user.groups.find_by(name: 'admin').present?
+  end
+
+  private
+
+  def categorize_rentals
+    @upcoming_rentals = @rentals.where('end_time > ? AND rental_status = ?', DateTime.current, 'reserved')
+    @ongoing_rentals = @rentals.where('rental_status = ?', 'checked_out')
+    @no_show_rentals = @rentals.where('end_time <= ? AND rental_status = ?', DateTime.current, 'reserved')
```
phantompatrol: Four(ish) critiques with these queries. We also did not do a sufficient job explaining who is using the index page, and for what reason.

1. Mea Culpa for not framing the use case. The index page will be used by Admins to view upcoming rentals for that day, or maybe today and tomorrow. Like a dashboard of customers who could be showing up, throughout the day. So, if we have a couple hundred reservations (on future date) those rentals will show up in `@upcoming_rentals`. Please amend your first query to look for rentals starting today, in the reserved status. HINT: You would be able to accomplish using the hash syntax `.where(key: value. key_two: value_two)`
2. I would strive for having the model pass the state values to you, as opposed to hard coding the strings. Double check the [gem documentation](https://github.com/aasm/aasm) to see if you can get the state names from method calls. Follow?
3. Double check to see if we are using similar lookups other places. If we are, I think this is a great target for a scope!
4. Using pure string conditions is not frowned upon, sometimes (like when using straight inequalities) you have to use it. However, for a variety of reasons [including security](http://guides.rubyonrails.org/active_record_querying.html#pure-string-conditions), using hash conditionals is much preferred by me. Further, you can often get away with using date arrays (for example), if you are trying to select a range. Lastly. remember if you ever get really stuck. there is not thing that prevents chaining `where`s as in: 'Model.where(key: value).where('date > ?', value)'

app/controllers/home_controller.rb
```diff
@@ -4,8 +4,25 @@ class HomeController < ApplicationController
...
+    @q = @rentals.where('rental_status = ?', 'checked_in').search(params[:q])
+    @past_rentals = @q.result(distinct: true).paginate(page: params[:page], per_page: @per_page)
+  end
+
+  def non_admin_filter
+    @upcoming_rentals = @upcoming_rentals.where('user_id = ?', @current_user.id)
+    @ongoing_rentals = @ongoing_rentals.where('user_id = ?', @current_user.id)
+    @no_show_rentals = @no_show_rentals.where('user_id = ?', @current_user.id)
```
phantompatrol: Begging for a scope!

phantompatrol:
Please add specs to the controller. Further, by moving some of your logic into scopes, you can move some of the burden to the model, where the shared functionality is more accessible and conceptually easier to test.

phantompatrol:
Lookin' pretty great as a first! Great description, photos, and use of partials

cmednieks:
Changed view so admins can see upcoming rentals within the next 24 hours on the home (all rentals) tab, and clicking on the 'future' tab will render all future rentals regardless of date

phantompatrol:
@cmednieks , you have a merge conflict in your `coverage/.last_run.json`

app/models/rental.rb
```diff
@@ -26,6 +26,11 @@ class Rental < ActiveRecord::Base
   alias_attribute :start_date, :start_time
   alias_attribute :end_date, :end_time
 
+  scope :upcoming_rentals, -> { reserved.where('start_time <= ? AND end_time >= ?', DateTime.current + 1.day, DateTime.current) }
+  scope :all_future_rentals, -> { reserved.where('end_time >= ?', DateTime.current) }
+  scope :no_show_rentals, -> { reserved.where('end_time < ?', DateTime.current) }
+  scope :users_rentals, -> (current_user_id) { where(user_id: current_user_id) }
```
dfaulken: Since you have

```ruby
belongs_to :user
```

in this model file, rather than a scope you could just write `some_user.rentals`.

spec/controllers/home_controller_spec.rb
```diff
@@ -3,29 +3,84 @@
...
-      get :index
-      expect(response).to render_template :index
+
+    describe 'for non_admin user' do
+      before(:each) do
+        @user = current_user
+        @other_user = create(:admin_user)
+        @user_rental = create(:mock_rental, user_id: @user.id )
+        @user_rental2 = create(:ongoing_rental, user_id: @user.id)
+        @not_users_rental = create(:mock_rental, user_id: @other_user.id)
```
phantompatrol: Hate to keep picking, but this could be simplified by passing the object as opposed to the id.

`create(:mock_rental, user_id: @user.id )` becomes `create(:mock_rental, user: @user)`

phantompatrol:
Lastly, please remove the two `.DS_Store` files from this PR. You will have to add .DS_Store to your global gitignore, then delete the two files

Issue #104, opened 2016-07-19
Small Issue with Rental Date range


owlee:
Date picker in rentals/new is not inclusive.
e.g. 
start_date 11/10/16
end_date 11/15/16

Users will expect reservation to persist from the 10th to the 15th, however, the code side only takes 10th to 14th. 

Plan
A. take a look at date picker options
B. modify JS
C. else Ruby parse to include actual end_date

** maybe for new programmers
** If no one gets to this before I do, I will pick up and solve.

PR #114, opened 2016-07-20
Hotfix for Rental Date Range Issue

Cobchise:
### Issue #104 ###

Small change to correct display for rental date range in calendar

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ gem 'date_validator'
 # gem 'capistrano-rvm'
 # gem 'capistrano-passenger'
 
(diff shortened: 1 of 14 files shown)


phantompatrol:
I am not sure this deals with the issue correctly, but I might also be lost in the 'off by one' sauce.

A reservation made by the APP from date A to date B should make a reservation in the API from date A to date B, and the dates being returned from the API should span from A to B. Inclusive in all cases.

phantompatrol:
Critique: A series of RSPEC tests would help developers down the road by elucidating the code and ensuring the correctness and completeness of the solution.

PR #115, opened 2016-07-22
Fix to inclusion of rental date issue

Cobchise:
### Issue #104 ###

- Added Display of Start/End Date to rental show
- Created Rentals will now be seen in database with correct start/end time in UTC
- FullCalendar will now display correct date range for rentals (inclusive)

> http://fullcalendar.io/docs/event_data/Event_Object/
EndParam: It is the moment immediately after the event has ended. For example, if the last full day of an event is Thursday, the exclusive end of the event will be 00:00:00 on Friday!

### DIFF
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -75,7 +75,6 @@ class RentalsController < ApplicationController
     @rental = Rental.new(rental_params)
 
     @start_date = params['start_date'] ? params['start_date'] : Time.zone.today
-
     if @rental.save
       flash[:success] = 'You have succesfully reserved your Rental!'
       redirect_to(@rental)
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -120,6 +120,7 @@ class RentalsController < ApplicationController
   # Only allow a trusted parameter "white list" through.
   def rental_params
     user = User.find(params.require(:rental).permit(:user_id)[:user_id])
-    params.require(:rental).permit(:start_time, :end_time, :item_type_id, :user_id).merge(department_id: user.department_id)
+    new_time = Time.zone.parse(params[:rental][:end_time]).end_of_day
+    params.require(:rental).permit(:start_time, :item_type_id, :user_id).merge(department_id: user.department_id, end_time: new_time)
   end
 end
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -3,6 +3,7 @@ class RentalsController < ApplicationController
 
   before_action :set_rental, only: [:show, :edit, :update, :destroy, :transform, :transaction_detail]
   before_action :set_item_types, only: [:index, :new, :create, :edit, :update, :processing]
+  before_action :set_items, only: [:index, :new, :create, :edit, :update, :processing]
   before_action :set_users, only: [:index, :new, :processing, :transform]
   before_action :set_incidental_types, only: [:new]
 
@@ -109,6 +110,10 @@ class RentalsController < ApplicationController
     @item_types = @item_types.where(name: params['item_type']).order(name: :asc) if params['item_type']
   end
 
+  def set_items
+    @items = Item.all
+  end
+
   def set_users
     @users = User.all
   end
@@ -121,6 +126,6 @@ class RentalsController < ApplicationController
   def rental_params
     user = User.find(params.require(:rental).permit(:user_id)[:user_id])
     new_time = Time.zone.parse(params[:rental][:end_time]).end_of_day
-    params.require(:rental).permit(:start_time, :item_type_id, :user_id).merge(department_id: user.department_id, end_time: new_time)
+    params.require(:rental).permit(:start_time, :item_type_id, :item_id, :user_id).merge(department_id: user.department_id, end_time: new_time)
   end
 end
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -1,7 +1,7 @@
 class RentalsController < ApplicationController
   @per_page = 10
 
-  before_action :set_rental, only: [:show, :edit, :update, :destroy, :transform]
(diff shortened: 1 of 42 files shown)


app/controllers/rentals_controller.rb
```diff
@@ -71,6 +71,7 @@ def rental_schedule
 
   # POST /rentals
   def create
+    params[:rental][:end_time] = Time.zone.parse(params[:rental][:end_time]).end_of_day
```
phantompatrol: Rather than rewriting the params, I would suggest saving it to a local variable

app/models/rental.rb
```diff
@@ -112,8 +112,8 @@ def event_status_color
   def self.to_json_reservations
     arr = all.each_with_object([]) do |rental, list|
       list << { title: rental.event_name,
-                start: rental.start_time.to_date,
-                end: rental.end_time.to_date,
+                start: rental.start_time,
+                end: rental.end_time,
```
phantompatrol: The red is soooo right here!

app/controllers/rentals_controller.rb
```diff
@@ -71,6 +71,7 @@ def rental_schedule
 
   # POST /rentals
   def create
```
phantompatrol: I know you stepped into a section of code that is untested, but if you could address that, we would very much appreciate it!

phantompatrol:
I know you stepped into a section of code that is untested, but if you could address that by adding some tests for the create action, we would very much appreciate it!

phantompatrol:
Looking good!

spec/factories/rental.rb
```diff
@@ -9,7 +9,12 @@
   end
 
   factory :invalid_rental, parent: :rental do
+    user_id { create(:user).id }
```
phantompatrol: You can clean this up by using `association`, as opposed to IDs. Because you already have access to a user factory, this can be written as:

`association :user`

Documentation: https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md

spec/factories/rental.rb
```diff
@@ -9,7 +9,12 @@
...
+    department_id { create(:department).id }
+    item_type_id { create(:item_type).id }
+    item_id { create(:item).id }
```
phantompatrol: Ditto 3x

spec/factories/rental.rb
```diff
@@ -9,7 +9,12 @@
...
     start_time nil
+    end_time (Time.current + 1.day).to_s
```
phantompatrol: No need to cast this to a string, it should be able to take a `Time` object.

spec/factories/rental.rb
```diff
@@ -30,4 +35,13 @@
...
   end
+
+  factory :create_rental, parent: :rental do
+    user_id { create(:user).id }
+    department_id { create(:department).id }
+    item_type_id { create(:item_type).id }
+    item_id { create(:item).id }
+    start_time Time.current.to_s
+    end_time (Time.current + 1.day).to_s
+  end
```
phantompatrol: You already have a `rental` factor, not sure why you need this section of code.

app/models/rental.rb
```diff
@@ -130,7 +130,7 @@ def sum_amount
   attr_accessor :skip_reservation_validation
 
   def create_financial_transaction
-    rental_amount = (((end_time.to_date - start_time.to_date).to_i - 1) * item_type.fee_per_day) + item_type.base_fee
+    rental_amount = ((end_time.to_date - start_time.to_date).to_i * item_type.fee_per_day) + item_type.base_fee
```
phantompatrol: I would love to see some tests of the rental model that ensures the correctness of your change.

A (hypothetical) zero day rental = 0? Error?
A one day rental with a base fee of `A` and a per day fee of `B` = X rental cost
A two day rental with a base fee of `A` and a per day fee of `B` = Y rental cost

spec/models/financial_transaction_spec.rb
```diff
@@ -35,7 +35,7 @@
       expect(transaction).to eq(polymorphism_trans)
       expect(rent).to eq(polymorphism_trans.rental)
 
-      expect(transaction.amount).to eq((((rent.end_time.to_date - rent.start_time.to_date).to_i-1)*rent.item_type.fee_per_day)+rent.item_type.base_fee)
+      expect(transaction.amount).to eq(((rent.end_time.to_date - rent.start_time.to_date).to_i*rent.item_type.fee_per_day)+rent.item_type.base_fee)
```
phantompatrol: Not a fan of the change, but the critique has more to due with this spec, rather than your work. So, separate issue tangentially related to your work but out of scope for the time being. https://github.com/umts/probable-engine/issues/117

spec/controllers/rentals_controller_spec.rb
```diff
@@ -60,6 +60,32 @@
...
+        expect do
+          post :create, rental: attributes_for(:invalid_rental)
+        end.to_not change(Rental, :count)
+      end
+      it 're-renders the :new template' do
+        post :create, rental: attributes_for(:invalid_rental)
+        expect(response).to render_template :new
+      end
+    end
+  end
```
phantompatrol: TY!

phantompatrol:
Last CR, I swear.

spec/models/rental_spec.rb
```diff
@@ -153,5 +150,26 @@
...
     expect(rental.financial_transaction).to be_an_instance_of(FinancialTransaction)
   end
+  it "doesn't allow a zero day rental" do
+    time = Time.current
+    rent = build(:mock_rental, start_time: time, end_time: time)
+    expect(rent).not_to be_valid
+  end
+  it 'creates a 1 day financial transaction with value: 100' do
+    rent = create :mock_rental, end_time: (Time.current + 1.second)
+    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([100])
```
phantompatrol: Ex post facto review: Could have used `find_by`

PR #109, opened 2016-07-20
Issue #95 - Added way for staff to uniquely identify vehicles

Cobchise:
#### Features: ####
1. **Scaffold** for items i.e. carts
2. Item model **delegates** to item_type model for some properties
2. Actions and views for **creating new carts** independent of model in API which are then pulled to local db
2. Ability to **edit** existing carts
3. **Implemented cart names throughout probable-engine** including scheduling rental etc
3. **Migrated Items into database** with unique name, item_type_id, and uuid
4. Implemented **'permanent-records'** for **'soft-deletion'** of items

PR #118, opened 2016-07-26
Rental scheduler

owlee:
Splitting of single and multi day reservation and getting rid of the simple_calendar for a much simpler form for clients.


app/assets/javascripts/global.js
```diff
@@ -42,7 +41,13 @@ function toggle_loading(){
 //Toggle a forms submit button when a checkbox is checked/unchecked
 function accept_and_enable(box)
 {
-    var submit = document.getElementById("TOC");
+    //var submit = document.getElementById("TOC");
+
+
+    var submit = $(box).parent().parent().parent().parent().find("input[type='submit']")[0];
```
phantompatrol: Did you (by any chance) find a better way to handle this? Maybe `parent_until` wth an additional `parent`, or `closest`?

app/views/rentals/_form.html.erb
```diff
@@ -1,64 +1,149 @@
...
+
+      <div class="form-group col-xs-12">
+        <div class="col-xs-offset-2 col-xs-6">
+          <div class="panel panel-primary">
+            <div class="panel-heading">
+              Terms and Conditions
+            </div>
+            <% @item_types.each do |type| %>
+              <div class="panel-body no-gutter" style="max-height: 10;overflow-y: scroll;">
+                <% hidden = '' %>
```
phantompatrol: ?

app/views/rentals/_form.html.erb
```diff
@@ -1,64 +1,149 @@
...
+                <% hidden = 'display:none; ' unless type == @item_types.first %>
```
phantompatrol: Could these 'hidden's be done with with a CSS class?

app/assets/javascripts/global.js
```diff
@@ -42,7 +41,13 @@ function toggle_loading(){
...
+    //console.log($(box).parentsUntil("#disclaimer_id").last()[0]);
```
phantompatrol: old comment 🐚 💨

app/assets/javascripts/global.js
```diff
@@ -42,7 +41,13 @@ function toggle_loading(){
 //Toggle a forms submit button when a checkbox is checked/unchecked
 function accept_and_enable(box)
 {
-    var submit = document.getElementById("TOC");
+    //var submit = document.getElementById("TOC");
```
phantompatrol: old comment 🍂

app/views/rentals/_form.html.erb
```diff
@@ -1,64 +1,149 @@
...
-  <div class="form-group col-xs-12">
-    <%= f.label :start_time, 'Start Date', class: 'control-label col-xs-2' %>
-    <div class="col-xs-6">
-      <%= f.date_field :start_time, value: @start_date, required: true, class: "form-control" %>
-    </div>
-  </div>
+      <div class="form-group col-xs-12">
+        <%= f.label :name, 'Cart Name', class: 'control-label col-xs-2' %>
+        <div class="col-xs-6">
+          <%= f.select :item_id, options_from_collection_for_select(@items.where(item_type_id: params["item_type_id"], deleted_at: nil).order(name: :asc), "id", "name"), {}, class: "selectpicker form-control" %>
```
phantompatrol: The `deleted_at: nil` should not be needed here. Could you verify, please

phantompatrol: Recommend moving the query up to the controller, for space and MVC reasons

app/views/rentals/_form.html.erb
```diff
@@ -1,64 +1,149 @@
...
-  <div class="form-group col-xs-12">
-    <%= f.label :end_time, 'End Date', class: 'control-label col-xs-2' %>
-    <div class="col-xs-6">
-      <%= f.date_field :end_time, value: @start_date.next_day, required: true, class: "form-control" %>
-    </div>
-  </div>
+      <div class="form-group col-xs-12">
+        <%= f.label :user, 'User', class: 'control-label col-xs-2' %>
+        <div class="col-xs-6">
+          <%= f.select :user_id, options_from_collection_for_select(@users.order(last_name: :asc), "id", "full_name"), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)" %>
```
phantompatrol: Ditto: https://github.com/umts/probable-engine/pull/118/files#r72329622

app/views/rentals/_form.html.erb
```diff
@@ -1,64 +1,149 @@
...
+        <%= f.label :type, 'Cart Type', class: 'control-label col-xs-2' %>
+        <div class="col-xs-6">
+          <%= f.select :item_type_id, options_from_collection_for_select(@item_types, "id", "name"), {}, class: "selectpicker form-control" %>
+        </div>
+      </div>
+
+      <div class="form-group col-xs-12">
+        <%= f.label :name, 'Cart Name', class: 'control-label col-xs-2' %>
+        <div class="col-xs-6">
+          <%= f.select :item_id, options_from_collection_for_select(@items.where(item_type_id: params["item_type_id"], deleted_at: nil).order(name: :asc), "id", "name"), {}, class: "selectpicker form-control" %>
```
phantompatrol: thrice: https://github.com/umts/probable-engine/pull/118/files#r72329622

app/views/rentals/_form.html.erb
owlee: I will look into it more. It was previously there before I made changes.

app/views/rentals/_form.html.erb
owlee: again ^^. I believe you can't from first look, it makes use of ruby functionality to determine display:none or not.

app/assets/javascripts/global.js
owlee: It's a seashell farting ha.

app/views/rentals/_form.html.erb
owlee: This piece of code stems from @Cobchise. I remember you mentioned it to him. Is there a fix in another PR? @phantompatrol

Cobchise: I added a default scope in PR #109 and changed the code in items_controller but I overlooked this code, that's my fault sorry about that. 
@owlee @phantompatrol 

> https://github.com/Cobchise/probable-engine/blob/4e130c394adfea2ef4b0c45a026e28fa5d1dd296/app/models/item.rb

phantompatrol: Whoops, that missed my CR. I created a separate issue to deal with this https://github.com/umts/probable-engine/issues/119

phantompatrol: tumblr_me97vzpj1o1ri63dso2_400

app/views/rentals/_form.html.erb
phantompatrol: So, maybe we were talking about two different ways of accomplishing the same thing. Granted, you are reusing someone else's code, but reusing code presents a great opportunity to refactor. Might also be why my issues take forever to close...

b1pdu


Anyway, back to the task at hand! Aside from the remaining inline style (which could be factored out), I was suggesting adding a class to hide the elements.

```SASS
.hidden-element
  display: none
```

```RUBY
<%= content_tag :span,
          type.disclaimer,
          class: "disclaimer hidden-element",
          id: "disclaimer_#{type.id}" %>
```

The advantage to moving the style from inline up to a rental or global SASS/CSS file is that we can change the style once, and change the whole site! Further, the JS is just as easy, as it turns into an issue of adding and removing classes on the element.

Side note, this might be a bug, but changing the item type does not hide/unhide the disclaimer (in the previous implementation). Am I missing is, or was this something that was never implemented?

phantompatrol: Additional side note, this is a naive proof of concept. The `display: none` could be moved into the style definition of disclaimer, and could be further refactored to be something like `disclaimer-hidden` vs `disclaimer-visable`.

Issue #93, opened 2016-07-14
As an admin, I should be able to process a vehicle that was not 'checked out'


phantompatrol:
It is conceivable that a customer may reserve a vehicle but 'no-show' for the pickup, and not cancel it from their end.

In this case, we need a separate method within the statemachine that moves the rental from 'reserved' to canceled, but also applies a possible financial transaction, email, etc.

It will function similar to https://github.com/umts/probable-engine/blob/master/app/models/rental.rb#L36, but execute an additional action.

Additional TODO: Get clarification from Jon as to the policy, if this were to happen.

PR #121, opened 2016-07-27
Processing no shows

cmednieks:
Regarding #93 , handles the case in which a customer makes a reservation and does not pick it up or cancel it prior to the end date:
- When 'process' is clicked in the home view, a form similar to check_in/out is rendered
- This form asks the admin if the customer was informed of their 'no-show' status (maybe use a mailer for this?)
- If yes, then rental status goes from reserved to canceled
- According to Jon, the customer will have already been charged for the rental and paid as a confirmation so there will be no need for further financial transactions
- Added quick fix for #125


app/controllers/rentals_controller.rb
```diff
@@ -127,4 +130,8 @@ def rental_params
     new_time = Time.zone.parse(params[:rental][:end_time]).end_of_day
     params.require(:rental).permit(:start_time, :item_type_id, :item_id, :user_id).merge(department_id: user.department_id, end_time: new_time)
   end
+
+  def sig_image_params
+    params.require(:rental).permit(:customer_signature_image)
+  end
```
phantompatrol: 👍

app/views/rentals/no_show_form.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <th>Responsible Person</th>
+    <th>Spire ID</th>
+    <th>Department</th>
+    <th>Rental Start Date</th>
+    <th>Rental End Date</th>
+  </tr>
+  <tr>
+    <td><%= rental.item_type.name%></td>
+    <td><%= rental.item.name%></td>
+    <td>$00</td>
```
phantompatrol: $00?

app/views/rentals/no_show_form.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <div class="col-xs-offset-2 col-xs-10">
+      <%= f.submit 'Process No Show', class: "btn btn-primary" %>
+      <%= link_to 'Back', rentals_path, class: "btn btn-default" %>
+    </div>
+  </div>
+<% end %>
+
+<script>
+  $("[name='contacted-customer?']").bootstrapSwitch();
+</script>
```
phantompatrol: Please move this up to a JS file

app/controllers/rentals_controller.rb
```diff
@@ -41,9 +41,11 @@ def processing
 
   # GET /rentals/1/transform
   def transform
-    if @rental.rental_status == 'reserved'
+    if @rental.reserved? && @rental.end_date < DateTime.current
```
phantompatrol: Please add some time_cop checks into the controller specs.

One concern: Confirm that time zones don't cause you issues. The app is running in eastern time, I just would like to see a test confirming the validity of the check.

phantompatrol: One thought and Question for the @umts/parking-programmers: Should we offer the ability to flag a no_show after the beginning or the end of the rental?

owlee: Beginning? So that the reservation availability can be recycled.

app/models/rental.rb
```diff
@@ -29,6 +29,7 @@ class Rental < ActiveRecord::Base
   scope :upcoming_rentals, -> { reserved.where('start_time <= ? AND end_time >= ?', DateTime.current + 1.day, DateTime.current) }
   scope :all_future_rentals, -> { reserved.where('end_time >= ?', DateTime.current) }
   scope :no_show_rentals, -> { reserved.where('end_time < ?', DateTime.current) }
+  scope :inactive_rentals, -> { where('rental_status = ? OR rental_status = ?', 'canceled', 'checked_in') }
```
phantompatrol: Perhaps refactor `where(rental_status: [canceled, checked_in])`? And perhaps use the methods, if you have access to them, as opposed to strings

app/models/rental.rb
```diff
@@ -63,6 +64,13 @@ class Rental < ActiveRecord::Base
     event :process do
       transitions from: [:inspected, :canceled], to: :available
     end
+
+    event :process_no_show do
+      transitions from: :reserved, to: :canceled
+      after do
+        update(checked_in_at: Time.zone.now)
+      end
```
phantompatrol: Double check to see if this needs to be set. Technically, the customer never checked in.

spec/models/rental_spec.rb
```diff
@@ -119,6 +119,11 @@
       expect(@rental).to be_checked_in
     end
 
+    it 'is canceled after being processed as a no show' do
+      @rental.process_no_show!
+      expect(@rental).to be_canceled
```
phantompatrol: `be_canceled`... cool?

app/controllers/rentals_controller.rb
cmednieks: Based on what I understood from talking to Jon, the customer will have been charged prior to the rental, thus reserving the cart for the amount of days they wanted. Therefore a customer could conceivably have reserved a cart for multiple days, but not necessarily show up on the first day. Since they'd have already paid for these days, the no-show flag shouldn't occur 'til after the end date.

app/models/rental.rb
cmednieks: @phantompatrol , not sure if there's a better way than using the strings along with the above syntax. I have access to the state methods individually but can't find an 'or' method where the scope will choose between two states. i.e. I can call a scope with '{ canceled }' and it will return the canceled rentals but not '{ canceled || checked_in}'

Issue #103, opened 2016-07-19
Add validation for FinancialTransaction amount and adjustment


phantompatrol:
FinancialTransaction `#amount` and `#adjustment` should likely never be undefined. They should be present and numerical, but default to 0 is not an issue in my mind.

PR #132, opened 2016-08-01
Adding validations to financial transactions

cmednieks:
- Addresses #103 
- Adds presence validations to `#amount` and `#adjustment` and defaults them to zero.


phantompatrol:
Seems very straight forward!

PR #135, opened 2016-08-02
Application helper specs

phantompatrol:


spec/helpers/application_helper_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'rails_helper'
 
 describe ApplicationHelper do
+  include Rails.application.routes.url_helpers
```
phantompatrol: Prune, not needed or used. Invoked accidently

PR #136, opened 2016-08-02
Fixing financial transaction invalid migration

### LINKED ISSUE #133: Invalid migration
There is an issue with the most recent migration:

```Ruby
1x-vl912-128-119-15-178:probable-engine logan$ bundle exec rake db:migrate
== 20160801150013 AddValidationsToFinancialTransactions: migrating ============
-- change_column(:financial_transactions, :amount, :integer, {:null=>false, :default=>0})
   -> 0.0500s
-- change_column(:financial_transactions, :adjustment, :integer, {:null=>false, :default=>0})
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Invalid use of NULL value: ALTER TABLE `financial_transactions` CHANGE `adjustment` `adjustment` int(11) DEFAULT 0 NOT NULL/Users/logan/projects/probable-engine/db/migrate/20160801150013_add_validations_to_financial_transactions.rb:4:in `change'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `load'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `<main>'
ActiveRecord::StatementInvalid: Mysql2::Error: Invalid use of NULL value: ALTER TABLE `financial_transactions` CHANGE `adjustment` `adjustment` int(11) DEFAULT 0 NOT NULL
/Users/logan/projects/probable-engine/db/migrate/20160801150013_add_validations_to_financial_transactions.rb:4:in `change'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `load'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `<main>'
Mysql2::Error: Invalid use of NULL value
/Users/logan/projects/probable-engine/db/migrate/20160801150013_add_validations_to_financial_transactions.rb:4:in `change'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `load'
/Users/logan/.rbenv/versions/2.3.1/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

The issue is that the migration works when there is no data in the database, but not if there are existing records. While is not a fundamental issue (because we have not gone to production), it is worth knowing how to deal with it effectively.

cmednieks:
Fixes #133 
- It appears that `change_column_null` does work to change pre-existing records in the database.
- Also got rid of the model level presence validation because amount and adjustment will default to 0 if not entered


app/models/financial_transaction.rb
```diff
@@ -4,6 +4,4 @@ class FinancialTransaction < ActiveRecord::Base
   belongs_to :transactable, polymorphic: true
 
   validates :rental_id, presence: true
-  validates :amount, presence: true
-  validates :adjustment, presence: true
```
phantompatrol: I understand why you eliminated this during test, but I think keeping this is fine!

db/migrate/20160801150013_add_validations_to_financial_transactions.rb
```diff
@@ -1,6 +1,8 @@
 class AddValidationsToFinancialTransactions < ActiveRecord::Migration
   def change
-    change_column(:financial_transactions, :amount, :integer, null: false, default: 0)
-    change_column(:financial_transactions, :adjustment, :integer, null: false, default: 0)
+    change_column_null(:financial_transactions, :amount, false, 0)
+    change_column_null(:financial_transactions, :adjustment, false, 0)
+    change_column(:financial_transactions, :amount, :integer, default: 0)
+    change_column(:financial_transactions, :adjustment, :integer, default: 0)
```
phantompatrol: Works as promised! Unfortunately, anyone who has pulled down master will still have reset, but I think the exercise was worth it. If someone still has old `financial_transactions` (old to your previous PR) in their db, this migration will not cause an issue for them. Confirmed!

spec/models/financial_transaction_spec.rb
```diff
@@ -32,14 +32,6 @@
       expect(transaction.amount).to eq(base_fee + daily_fee)
     end
 
-    it 'is invalid with a nil amount' do
-      expect(build :financial_transaction, :with_rental, amount: nil).not_to be_valid 
-    end
-
-    it 'is invalid with a nil adjustment' do
-      expect(build :financial_transaction, :with_rental, adjustment: nil).not_to be_valid
-    end
```
phantompatrol: Again, I would love to keep these!

spec/models/financial_transaction_spec.rb
```diff
@@ -32,12 +32,12 @@
...
     end
 
-    it 'is invalid with a nil amount' do
+    it 'is invalid without an amount' do
       expect(build :financial_transaction, :with_rental, amount: nil).not_to be_valid 
     end
 
-    it 'is invalid with a nil adjustment' do
-      expect(build :financial_transaction, :with_rental, adjustment: nil).not_to be_valid
+    it 'is invalid without an adjustment' do
```
phantompatrol: What a difference in readability that makes!

phantompatrol:
Odd, that Travis error is not you. Checking!

phantompatrol:
https://github.com/umts/probable-engine/issues/139

PR #138, opened 2016-08-04
Increasing coverage

cmednieks:
Brings coverage for incidental types controller and incurred incidentals controller up to 100%


phantompatrol:
@cmednieks , There was a bit of a sea change between when this was submitted (like 30 mins ago) and our new rails 5 controller specs. See: https://github.com/Cobchise/probable-engine/blob/9e5650889307e211007464087cb2010680b917a0/spec/controllers/groups_controller_spec.rb for a couple of examples.

Long story short, this looks GREAT! It is just that controller specs have change a smidgen.

phantompatrol:
You should be able to `git pull origin master` to see the which files need to be adjusted. Sorry about!

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.38
+    "covered_percent": 99.71
```
phantompatrol: Holy ****. So Close.

PR #140, opened 2016-08-04
Addresses an issue with failing application helper specs, post rails 5 migration

phantompatrol:
* Adds route helper explicitly
* Sets a default URL for development and test


phantompatrol:
@cmednieks , this should fix your [PR](https://github.com/umts/probable-engine/pull/136). One moment while Travis runs

PR #146, opened 2016-08-09
fixed single day rental issue

### LINKED ISSUE #145: Error thrown when trying to create a 'single-day' rental any day other than today
When I try to create a 'single-day' rental on any day except the current day it throws an error. 

I believe this is because the rental is being created using a 'hidden-field' for the end date thats value is always the current date. So when the start date is changed to be after this set end date, the rental validations kick in and an error is thrown.

cmednieks:
Fixes #145 in controller, users can now create single day rentals in the future

edit: not sure if this will be used as others are working on a javascript fix for this. It should work for the upcoming milestone, however

### DIFF
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -135,7 +135,7 @@ class RentalsController < ApplicationController
     new_time = if params[:commit] == 'Create Single Day Rental'
                  Time.zone.parse(params[:rental][:start_time]).end_of_day
                else
-                 new_time = Time.zone.parse(params[:rental][:end_time]).end_of_day
+                 Time.zone.parse(params[:rental][:end_time]).end_of_day
                end
     new_time
   end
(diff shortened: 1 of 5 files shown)


phantompatrol:
RE: `edit: not sure if this will be used`, what do others think of this as a solution? @owlee @Cobchise @xliu3

Cobchise:
@phantompatrol @owlee @cmednieks @xliu3
I think both solutions have merits, the javascript being maybe a little cleaner with less code but the controller method maybe being a little more 'robust' in my opinion. 
I honestly have to say I personally prefer Charles' method just because its provides rspec testing, I like his assign_time method and I think it is a little more intuitive as we are already altering the end_time in the controller as is. 
I think if we are going to approach it from javascript I'd like to redo the method I PR'd as the current one relies on the javascript only finding the first element on the page, which I don't really like. I only really put that PR there for comparison purposes.

That's my gut feel but at the end of the day both methods work so..

phantompatrol:
Closing this, since we merged the other one?

PR #147, opened 2016-08-09
Added javascript for updating end time correctly on single day rental

### LINKED ISSUE #145: Error thrown when trying to create a 'single-day' rental any day other than today

Cobchise:
### Fixes #145 ###

### DIFF
--- a/app/assets/javascripts/global.js
+++ b/app/assets/javascripts/global.js
@@ -37,6 +37,11 @@ $(document).ready(function() {
         })
       })
     })
+
+    //Single day submit button function for updating end time
+    $('#TOC').click(function () {
+      $('#rental_end_time').val($('#rental_start_time').val())
+    })
 });
 
 //Toggle the loading div


phantompatrol:
RE: #146 , and this [comment](https://github.com/umts/probable-engine/pull/146#issuecomment-238839816), which one of these two solutions do we like best? @cmednieks @owlee @xliu3

owlee:
I am going to say JS now client sided vs server sided change in case we have multiple client instances for the same time slot. I haven't found a clean solution (esp with modifying the datepicker() because bootstrap overrides my JS func).

phantompatrol:
Having multiple clients is not an issue in either solution, as the reservation is being made as the rental saves.

I would just say to merge the one we like best, close the opposing one and cheer at merging your first PR

owlee:
👍  Do I get merging powers from you or Metin? @phantompatrol

phantompatrol:
@umayavuz !

owlee:
Can't wait to merge my first PR at code review! 😰

Issue #150, opened 2016-08-11
Process (early warning popup)


owlee:
As an admin I would like a popup to warn me if I am about to process a rental early.

- alert() on the transform page
- only popup if the rental is processing is earlier than the start date of the rental

PR #151, opened 2016-08-12
Adding popup warning for early checkout

cmednieks:
Per #150 and code review discussion, adds popup warning user that they are about to process an early checkout


phantompatrol:
Sure, seems simple enough.

Issue #154, opened 2016-08-22
Incorrect Rental Costs


cmednieks:
The rental amounts that appear on the app are inconsistent with the rental rate chart, shown below, forwarded to us by @phantompatrol a few weeks back. The discounts for long-term rentals (e.g. a 3-week, 4-seat cart rental would cost $50 less than expected from the base fee + daily fees) also are not currently implemented when creating the financial transaction.

![screen shot 2016-08-22 at 1 24 51 pm](https://cloud.githubusercontent.com/assets/19389086/17864479/42dff972-686c-11e6-8267-4d06a680c147.png)

phantompatrol:
You might want to check in with Jon, with respect to offering the long term rates to customers. I believe it is something that he wanted to set now and not offer, oddly enough.

josephkelley:
I spoke to Jon about this today. We will finalize numbers at a later date, but for now the plan is to have flat daily fees. Also, for every six days in a rental period, one day will be free.

For particularly long-term rentals and other cases where we would like to deviate from the flat expense scale, payment can be handled outside the application -- or maybe we could support negative incidentals?

PR #160, opened 2016-09-08
Fix golf cart pricing

josephkelley:
Update pricing to reflect flat rates called for in #154.


josephkelley:
Disregard the commits from @cmednieks. They are already in production, I just did a sloppy rebase when I was trying to figure out squashing.

Issue #108, opened 2016-07-20
Rental#delete_reservation Has some Untested Code


phantompatrol:
There is untested code in the delete_reservation callback method, for objects of type Rental. Please resolve that.

Issue #105, opened 2016-07-19
Rental#create_reservation Has Untested Code


phantompatrol:
Issue: Untested code in `create_reservation`

Please induce a set of reasonable errors (404, 402, 500, lack of availability, etc) to trigger the rescue in [create_reservation](https://github.com/umts/probable-engine/blob/master/app/models/rental.rb#L68).

Additional Story: As a user, I would like connection, permission, and server issues handled differently from a lack of availability. (Please inform the customer, if the rental fails, that it is due to a issue with the API or a lack of available carts)

Bonus points, refactor the method to use an implicit `begin`

PR #165, opened 2016-09-14
Rental #create_reservation and #delete_reservation tests

cmednieks:
Addresses coverage issues brought up in #108 and #105


spec/models/rental_spec.rb
```diff
@@ -223,4 +223,31 @@
     rent = create :mock_rental, item_type: create(:item_type, name: 'Test 220', base_fee: 200, fee_per_day: 20)
     expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([220])
   end
+
+  describe '#delete_reservation' do
+    context 'error thrown' do
+      it 'logs the error and returns false' do
+        r = build(:rental)
+        r.create_reservation
```
phantompatrol: Are the two lines needed, by any chance? Or are the callbacks not firing? What I am suggesting is something along the lines of...

```Ruby
it 'logs the error and returns false' do
  allow(Inventory).to receive(:delete_reservation).and_raise(InventoryExceptions::AuthError)
  expect(create(:rental).delete_reservation).to be false
end
```

spec/models/rental_spec.rb
```diff
@@ -223,4 +223,31 @@
...
+      it 'logs error and returns false for a series of errors' do
+        r = build(:rental)
+        allow(Inventory).to receive(:create_reservation).and_raise(InventoryExceptions::InventoryError)
+        expect(r.create_reservation).to be false
+        r = build(:rental)
+        allow(Inventory).to receive(:create_reservation).and_raise(InventoryExceptions::ReservationError)
+        expect(r.create_reservation).to be false
+        r = build(:rental)
+        allow(Inventory).to receive(:create_reservation).and_raise(InventoryExceptions::AuthError)
+        expect(r.create_reservation).to be false
```
phantompatrol: Ditto to the above comment

spec/models/rental_spec.rb
cmednieks: They do appear to be necessary; I tried to do the above code and it seems that building a rental does not trigger `create_reservation` . I've noticed while working with the unfriendly errors that the callbacks don't seem to be working the way that we've been expecting them to and am trying to fix that problem in an upcoming PR.

tomecho:
@cmednieks please resolve merge conflict with master

PR #166, opened 2016-09-14
Cancel rentals fix

### LINKED ISSUE #159: Cannot cancel rentals
After clicking 'cancel' on a rental, a green flash message appears saying that the rental has been canceled, but its state does not change.

tomecho:
fixes #159


Gemfile
```diff
@@ -66,7 +66,7 @@ gem 'date_validator'
 # permanent records
 gem 'permanent_records'
 
-gem 'aggressive_inventory', '0.0.2', git: 'https://github.com/umts/aggressive_inventory.git'
+gem 'aggressive_inventory', '0.0.2', git: 'https://github.com/umts/aggressive_inventory.git' # isnt version is ignored here? it should be 'tag'
```
phantompatrol: Yes, it is a holdover from including the gem locally. That is using `path:` instead of `git:`

app/views/rentals/_table.html.erb
```diff
@@ -20,7 +20,10 @@
       <td><%= number_to_currency(rental.financial_transaction.amount) %></td>
       <td class="min-width">
         <%= link_to (content_tag(:span, nil, class: "glyphicon glyphicon-eye-open")+" View"), rental, class: "btn btn-xs btn-success" %>
-        <%= link_to (content_tag(:span, nil, class: "glyphicon glyphicon-ban-circle")+" Cancel"), rental, method: :delete, data: { confirm: 'Are you sure?' }, class: "btn btn-xs btn-danger" unless rental.canceled? || rental.checked_in? %>
+        <% if rental.reserved? && !rentals.no_show_rentals.include?(rental) %>
```
phantompatrol: Perhaps: [`rentals.no_show_rentals.include?(rental)`](http://guides.rubyonrails.org/active_support_core_extensions.html#exclude-questionmark)

Gemfile
josephkelley: Either remove the comment or implement the change you mention within it : )

Issue #164, opened 2016-09-14
Rental Creation Non-UserFriendly Error


cmednieks:
When a new rental creation on the app is attempted and fails (due to lack of available carts in this instance) the 'new' page is not re-rendered with a warning message as expected. Instead, the following error page is rendered: 
![screen shot 2016-09-14 at 9 57 22 am](https://cloud.githubusercontent.com/assets/19389086/18515210/90330cca-7a62-11e6-9af7-c347bec4bdd9.png)

PR #167, opened 2016-09-14
Fixing issue with unfriendly error when rental creation fails

cmednieks:
This fixes the issue addressed in #164 
- Tells user if the item type they requested is not available. If this is not the issue, it gives a generic flash warning.

### DIFF
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -76,11 +76,16 @@ class RentalsController < ApplicationController
 
     @start_date = params['start_date'] || Time.zone.today
 
-    if @rental.save
+    if @rental.create_reservation && @rental.save
       flash[:success] = 'You have succesfully reserved your Rental!'
       redirect_to(@rental)
     else # error has problem, cannot rental a error message here
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
+      set_users
+      if @rental.item_type_id.present? && Rental.where(item_type_id: @rental.item_type_id).count == Item.where(item_type_id: @rental.item_type_id).count
+        flash[:danger] = "#{ItemType.find(@rental.item_type_id).name} is unavailable for this date."
+      else
+        flash[:danger] = 'Something went wrong. Please try again.'
+      end
       render :new
     end
   end
--- a/spec/controllers/rentals_controller_spec.rb
+++ b/spec/controllers/rentals_controller_spec.rb
@@ -86,7 +86,7 @@ describe RentalsController do
           post :create, params: { rental: invalid_create }
         end.to_not change(Rental, :count)
       end
-      it 're-renders the :new template' do
+      it 're-renders the :new template with a generic warning message' do
         post :create, params: { rental: invalid_create }
         expect(response).to render_template :new
       end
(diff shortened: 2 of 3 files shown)


app/controllers/rentals_controller.rb
```diff
@@ -76,11 +76,16 @@ def create
...
 
-    if @rental.save
+    if @rental.create_reservation && @rental.save
       flash[:success] = 'You have succesfully reserved your Rental!'
       redirect_to(@rental)
     else # error has problem, cannot rental a error message here
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
+      set_users
+      if @rental.item_type_id.present? && Rental.where(item_type_id: @rental.item_type_id).count == Item.where(item_type_id: @rental.item_type_id).count
+        flash[:danger] = "#{ItemType.find(@rental.item_type_id).name} is unavailable for this date."
```
cmednieks: Will add testing for this line of code on Monday

cmednieks:
Will add testing and fix codeclimate when I'm back on monday

app/controllers/rentals_controller.rb
```diff
@@ -76,11 +76,16 @@ def create
 
     @start_date = params['start_date'] || Time.zone.today
 
-    if @rental.save
+    if @rental.create_reservation && @rental.save
```
phantompatrol: Do we not have a hook to create the reservation automagically?

app/controllers/rentals_controller.rb
```diff
@@ -76,11 +76,16 @@ def create
...
       flash[:success] = 'You have succesfully reserved your Rental!'
       redirect_to(@rental)
     else # error has problem, cannot rental a error message here
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
+      set_users
+      if @rental.item_type_id.present? && Rental.where(item_type_id: @rental.item_type_id).count == Item.where(item_type_id: @rental.item_type_id).count
```
phantompatrol: 1. Perhaps you can shorten the syntax a bit: `if Rental.where(item_type: @rental.item_type).count == Item.where(item_type: @rental.item_type).count`. If Rentals or Items have a `nil` item_type, we are going to give ourselves headaches :-p
2. I am not sure this logic works without considering rentals over time. That is, rentals in the past, present, and future.

spec/controllers/rentals_controller_spec.rb
```diff
@@ -86,7 +86,7 @@
           post :create, params: { rental: invalid_create }
         end.to_not change(Rental, :count)
       end
-      it 're-renders the :new template' do
+      it 're-renders the :new template with a generic warning message' do
```
phantompatrol: Understood that this might be a work in progress, but some additional tests for the new logic would be a 👍

app/controllers/rentals_controller.rb
phantompatrol: Perhaps just: `@rental.item_type.name` or `@rental.item_type.try(:name)` if you are worried about `nil`

app/controllers/rentals_controller.rb
cmednieks: We do, but it's not checking that `create_reservation` works correctly. If `create_reservation` doesn't work and returns false, it will still attempt to perform `@rental.save` thus causing the error page to appear. Not sure of the best way besides this to verify that the hook runs correctly.

cmednieks:
closing this while i rethink the logic

Issue #184, opened 2016-09-26
Replace Check Out/Check In with Pick Up/ Drop Off


Cobchise:
Language Change

- Replace both user visible labels and code to avoid confusion
- Create migration to change database entries

cmednieks:
working on this

PR #186, opened 2016-09-26
Wording Changes

cmednieks:
changed check in/out to drop off/pick up as requested in #184


app/models/digital_signature.rb
```diff
@@ -3,5 +3,5 @@ class DigitalSignature < ActiveRecord::Base
   belongs_to :rental
   validates :image, :intent, :author, presence: true
   enum author: { csr: 0, customer: 1 }
-  enum intent: { check_out: 0, check_in: 1 } # later add intent damages
+  enum intent: { pickup: 0, drop_off: 1 } # later add intent damages
```
owlee: pick_up?

spec/factories/digital_signature.rb
```diff
@@ -2,7 +2,7 @@
   factory :digital_signature do
     association :rental, factory: :mock_rental
     image 'fake image'
-    intent :check_out
+    intent :pickup
```
owlee: refer above for consistency sake

PR #193, opened 2016-10-11
importing everything causes 404's

tomecho:


app/assets/stylesheets/application.css.sass
```diff
@@ -18,7 +18,7 @@
 @import "bootstrap-sprockets"
 @import "bootstrap"
 @import "bootstrap-datetimepicker"
-@import "./**/*" 
+//@import "./**/*"
```
owlee: Does this work?

isiryk: seems to yes

Issue #200, opened 2016-10-28
FinancialTransaction.adjustments


tomecho:
Should this field be used to modify the amount field, currently the cost of a financial transaction is just the amount, it doesnt look like this field is in use.  Should flexible pricing be useing this cost adjustment?

tomecho:
@phantompatrol @owlee

owlee:
It could be blown away or left alone. I don't see any harm either way, but, may or may not be useful in the future.

owlee:
![Alt Text](https://media.giphy.com/media/KVUhxQAqrAV3i/giphy.gif)

PR #192, opened 2016-10-11
Js verification in Rental Reservation

owlee:
Js to display the reservation costs.

Caution for merge because there are some weird errors and not sure if it is my branch.
- missing attachments. I will need to create some storage for it on the DB.

screen shot 2016-10-11 at 2 20 46 pm

Issue #153, opened 2016-08-18
Admins would like to visually see the number of dates per rental and the total cost


owlee:
-JS to show the date range and pricing under the submit
-hub/TPP for reference

owlee:
@owlee almost done

tomecho:
fixed by #215

PR #201, opened 2016-10-28
Flexible pricing

### LINKED ISSUE #181: Add flexible rental pricing
When a user initiates a rental, we will populate a new "fee" field with a dollar amount calculated by the existing formula. We would like this price to be editable by CSRs in order to handle extreme cases such as month-long rentals.

tomecho:
such ui
very ux
wow
resolves #181 
needs to be merged with #192 which is a fix for #153
and waiting on #200 to be answered


db/seeds.rb
```diff
@@ -33,6 +33,7 @@
 
   puts 'Creating Model Item Type'
   item_types = YAML::load_file(File.join(Rails.root, 'db/db_yml', 'item_types.yml'))
+  binding.pry
```
owlee: note

PR #202, opened 2016-10-31
change all back buttons

### LINKED ISSUE #197: Correct usage of back buttons
Some 'Back' buttons, such as in form pages, link to a specific page such as index or show pages instead of using the traditional browser back action as would be expected. 
Change all such instance to make more sense.

Back buttons should be encoded as:
`<%= link_to 'Back', :back, class: "btn btn-default" %>`

tomecho:
resolves #197

### DIFF
--- a/app/views/rentals/show.html.erb
+++ b/app/views/rentals/show.html.erb
@@ -117,11 +117,7 @@
 <div class="form-group col-xs-12">
   <div class="col-xs-offset-2 col-xs-10">
     <%#= link_to 'Edit Rental', edit_rental_path(@rental), class: "btn btn-primary" %>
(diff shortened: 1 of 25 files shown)


app/views/rentals/show.html.erb
```diff
@@ -117,11 +117,7 @@
 <div class="form-group col-xs-12">
   <div class="col-xs-offset-2 col-xs-10">
     <%#= link_to 'Edit Rental', edit_rental_path(@rental), class: "btn btn-primary" %>
-    <% if @current_user.has_permission?(:rentals, :index) %>
```
owlee: weird code...

PR #203, opened 2016-11-02
Friendlier errors

cmednieks:
Fixing #164 , It gets rid of the unfriendly error page when there are no more carts available by redirecting back to the :new rental form. It does so by getting rid of the hooks in the model and calls `create_reservation` and `delete_reservation` explicitly in the rental controller.


cmednieks:
Not sure why the merge did not go through correctly...

tomecho:
@cmednieks if you're still available, did you mean to remove fee schedules from schema?

cmednieks:
@tomecho oops, did not mean to do that

tomecho:
#206 made some edits and put into

PR #213, opened 2016-11-22
Enhanced user selection

### LINKED ISSUE #171: enhanced user selection
on reservation#new the user should be searchable by name, spire, etc.

### LINKED ISSUE #212: Rental's should track their creator
* User_id is now for customer, we should still be tracking who created the rental.
* user_id should be renamed to renter_id
* creator_id will be the other id which is the user that created that rental (which should just be `@current_user`)

tomecho:
resolves #171 
resolves #212 

review this at the tuesday code review


app/views/rentals/_search_users_table.html.erb
```diff
@@ -0,0 +1,26 @@
+<%= will_paginate @users, renderer: BootstrapPagination::Rails %>
+<table class="table table-striped table-bordered table-hover">
+  <thead>
+    <tr>
+      <th>First name</th>
+      <th>Last name</th>
+      <th>Email</th>
+      <th>Spire</th>
+      <th>Department</th>
+      <th class="min-width">Actions</th>
```
dfaulken: What's a class of min-width do?

app/views/rentals/_search_users_table.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @users.each do |user| %>
+      <tr>
+        <td><%= user.first_name %></td>
+        <td><%= user.last_name %></td>
+        <td><%= user.email %></td>
+        <td><%= user.spire_id %></td>
+        <td><%= user.department.try(:name) %></td>
```
dfaulken: Is this association validated? If not, should it be?

db/seeds.rb
```diff
@@ -29,8 +29,10 @@
 
   puts 'Putting users in Parking department'
   parking = Department.find_by name: 'Parking'
-  parking.users << User.all
-
+  parking.users << User.all[0...((User.count/2).floor)]
```
dfaulken: How about

```ruby
parking.users, transit.users = User.all.each_slice(User.count / 2).to_a
```

spec/controllers/home_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:future) { create :far_future_rental, reservation_id: 7777 }
   let(:ongoing) { create :ongoing_rental, reservation_id: 8888 }
   let(:ongoing2) { create :ongoing_rental, reservation_id: 9999 }
-  let(:canceled) { create :mock_rental, rental_status: 'canceled', reservation_id: 0000 }
+  let(:canceled) { create :mock_rental, rental_status: 'canceled', reservation_id: 0o000 }
```
dfaulken: wat

spec/controllers/home_controller_spec.rb
```diff
@@ -74,8 +74,8 @@
 
       it 'filters out rentals not belonging to the current user' do
         get :index
-        expect(assigns[:rentals]).to include(create(:mock_rental, user_id: @user.id), create(:ongoing_rental, user_id: @user.id))
-        expect(assigns[:rentals]).not_to include(create(:mock_rental, user_id: @other_user.id))
+        expect(assigns[:rentals]).to include(create(:mock_rental, renter_id: @user.id), create(:ongoing_rental, renter_id: @user.id))
```
dfaulken: Here too, don't need to assign by ID:

```ruby
create :mock_rental, renter: @user
```

dfaulken: Also, why does your factory have the word `mock` in it? Aren't they all mock things in a test?

spec/controllers/rentals_controller_spec.rb
```diff
@@ -244,4 +244,35 @@
...
+
+  describe 'GET #search_users' do
+    before(:each) do
+      User.destroy_all
+    end
+
+    it 'returns everything if cant find a user' do # dynamically searches by email, spire, fullname and department (in that order)
+      create_list :user, 8 # can only do 8 because of pagination
+      get :search_users, params: { user_search_query: '$%%!$#' } # this query should return no users
+      expect(assigns[:users]).to eq(User.all.to_a)
```
dfaulken: Yeah, I still disagree with this.

spec/controllers/rentals_controller_spec.rb
```diff
@@ -244,4 +244,35 @@
...
+    end
+
+    it 'finds by spire' do
+      this_one = create :user, spire_id: 86_753_091 # rubocop demanded i separate this huggeeee number with underscores (which are ignored)
+      get :search_users, params: { user_search_query: '86753091' } # should be unique enough
+      expect(assigns[:users]).to eq([this_one]) # will be exact match
```
dfaulken: BTdubs, there's a `contain_exactly` matcher.

app/controllers/home_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
     @rentals = if @current_user.groups.where(name: 'admin').present?
                  Rental.all
                else
-                 @current_user.rentals
+                 Rental.where(renter: @current_user)
```
dfaulken: Now that I understand why, I'd recommend a scope so that you could say something like

```ruby
Rental.rented_by @current_user
```

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -55,10 +54,53 @@ function costCalculation(){
...
+    },
+    error: function () {
+      alert("Failed to retreive users.");
+    }
+  });
+}
+
+function modalSearchResultsFormatPaginationLinks(undefined, link) {
+  link = $(link); //jquery that
+  var pageNo = link.attr("href").split("&")[0].split("?")[1].split("=")[1]; //works universially to pull pageNo out of link
```
tomecho: unfortunately this is a very specific thing, using will_paginate within a partial retrieved by an ajax request.  I searched but was unable to find an existing library.

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -69,4 +111,16 @@ $(document).ready( function () {
   $('select.cost-dependent').on('change', function () {
     costCalculation();
   });
+
+  // ~ user searching
```
tomecho: not at all, i was just commenting that the hooks below are all for user searching.

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -69,4 +111,16 @@ $(document).ready( function () {
...
+
+  //enter key performs click on search for ease of use
+  $(".userSearchQuery").keypress(function (key) {
+    if(key.which === 13) {
```
tomecho: ascii enter key, i wrote that in line 117

app/assets/javascripts/rentals/new.js.erb
dfaulken: But the parameter you're pulling out of there has a name, right?

app/assets/javascripts/rentals/new.js.erb
dfaulken: OK.

app/assets/javascripts/rentals/new.js.erb
dfaulken: Ah, okay.

app/controllers/home_controller.rb
tomecho: not any more it will not, `user.rentals` will retrieve a list of all associated rentals for that user including ones they reserved and the ones that they are the renter.  There are two fields in `Rental` that refer to users, `renter_id` and `creator_id`.  The is often different than the renter, for example when a csr creates it.

app/controllers/rentals_controller.rb
```diff
@@ -22,6 +23,27 @@ def index
...
   end
 
+  # GET /rentals/search_users?q
+  def search_users
+    # prioritiezed by order queries
+    query_priority = %i(email_cont spire_id_eq full_name_cont)
+    @users = []
+
+    query_priority.each do |q|
+      @users += User.ransack(q => params[:user_search_query]).result
```
tomecho: I do not think so.

app/controllers/home_controller.rb
dfaulken: Right, see my second comment. I wrote that, and then later read the `Rental` model file.

dfaulken: Also, you removed `User.has_many :rentals`, so I think `user.rentals` doesn't work anymore.

app/controllers/rentals_controller.rb
```diff
@@ -22,6 +23,27 @@ def index
...
+      # break if @users.present? # maybe just join all the lists
+    end
+
+    # could be a department too
+    ## join users with departments
+    @users += Department.ransack(name_cont: params[:user_search_query]).result.map(&:users).flatten
+
+    @users = User.all if @users.blank? # if we couldnt find anything pass them everything
```
tomecho: No, feedback, willfix

dfaulken: Personally, it seems more straightforward to return a blank collection. Then in the view you can simply say 

```haml
- if @users.blank?
  Your search returned no results.
```

app/controllers/rentals_controller.rb
```diff
@@ -22,6 +23,27 @@ def index
...
+
+    @users = @users.uniq.paginate(page: params[:page], per_page: 8) # remove duplicates and split into pages
```
tomecho: Like on line 33?
It needs to be done in one centralize location afterwards because it will search by all types.
If i have a User(first_name: "jim", email: "jim@gmail.com"), I will match the same user twice.

dfaulken: Odd. Ransack doesn't have a way of searching by multiple attributes at once?

app/views/rentals/_search_users_table.html.erb
tomecho: sets minimum width of a column, its a standard across the application.

dfaulken: Ref: https://github.com/umts/probable-engine/blob/master/app/assets/stylesheets/tables.css#L21-L24

So I'm still confused though. Why do you want a 1px-wide table column?

app/views/rentals/_search_users_table.html.erb
tomecho: not validated, not all users have departments.

db/seeds.rb
tomecho: that's a much better approach thanks!

spec/controllers/home_controller_spec.rb
tomecho: rubocop demanded i place this in there because the number was far too many digits to be readable.

tomecho: perhaps I should say, too many consecutive same digits.

dfaulken: ...but why is this an octal number at all?

app/assets/javascripts/rentals/new.js.erb
tomecho: yes it does.

app/controllers/rentals_controller.rb
tomecho: The point is that this dynamic search field will match more than one field, I only want to search by one attribute.

spec/controllers/home_controller_spec.rb
tomecho: I believe the designation is for a soft rental, one that does not reach out the api and a persistent rental that does contact the api.

app/assets/javascripts/rentals/new.js.erb
dfaulken: Then it stands to reason that you can use a queryString library to pull it out by name.

app/views/rentals/_search_users_table.html.erb
tomecho: dunno its just the standard, its even generated by the rails generator

dfaulken: Okay, I won't worry about it then

app/controllers/rentals_controller.rb
dfaulken: Didn't you give an example above of searching by multiple attributes?

db/seeds.rb
tomecho: actually this seems to only take equal halves, there are 9 users, this method takes 4, puts them in parking users, 4 more and puts them in transit users.  Albeit I think this is fine because users are not required to have a department, leaving one without a department here is fine.

tomecho:
ahhh is that what you mean?

--Tom
_________________________________________

Thomas Arthur Peck IV
NETID: tapeck
Email: tapeck@umass.edu
Spire ID: 28950694

On Wed, Jan 4, 2017 at 4:06 PM, David Faulkenberry <notifications@github.com

app/controllers/rentals_controller.rb
tomecho: ohhh i see what you mean by multiple attributes, i thought you ment i wanted to search for them all at the same time like with the query `test` it would contain it for each match.

db/migrate/20161128184121_add_creator_id_and_renter_id_to_rentals.rb
```diff
@@ -0,0 +1,7 @@
+class AddCreatorIdAndRenterIdToRentals < ActiveRecord::Migration[5.0]
+  def change
+    add_column :rentals, :creator_id, :integer
+    add_column :rentals, :renter_id, :integer
+    remove_column :rentals, :user_id # no longer needed
```
dfaulken: Specifying a column type is optional here, but I think it's preferable in general, so that this migration could be rolled back if needed.

db/seeds.rb
```diff
@@ -92,7 +93,7 @@
 puts 'Updating permissions'
 admin = Group.find_by name: 'admin'
 Permission.update_permissions_table
-Permission.create(controller: 'rentals', action: 'cost_adjustment')
+Permission.create(controller: 'rentals', action: 'cost_adjustment') # this is a special permission that isnt really a controller action
```
dfaulken: ... wat?

spec/controllers/home_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:future) { create :far_future_rental, reservation_id: 7777 }
   let(:ongoing) { create :ongoing_rental, reservation_id: 8888 }
   let(:ongoing2) { create :ongoing_rental, reservation_id: 9999 }
-  let(:canceled) { create :mock_rental, rental_status: 'canceled', reservation_id: 0000 }
+  let(:canceled) { create :mock_rental, rental_status: 'canceled', reservation_id: 00_00 }
```
dfaulken: So I still have my question about why you need four zeros here.

app/controllers/rentals_controller.rb
```diff
@@ -22,6 +23,28 @@ def index
   def show
   end
 
+  # GET /rentals/search_users?q
+  def search_users
+    @users = []
+
+    if params[:user_search_query].present?
+      # prioritiezed by order queries
```
dfaulken: typo

spec/controllers/home_controller_spec.rb
tomecho: I think someone was trying to keep it standardized with the 4 character id. The reason they are set is because the factory is improperly configured, each rental factory doesnt have a shared fake reservation_id

db/seeds.rb
tomecho: this permission allows the user the adjust prices

spec/controllers/home_controller_spec.rb
dfaulken: 4 characters? Aren't they numbers?

db/seeds.rb
dfaulken: ... that isn't a route? How do you use it then?

app/models/user.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
+include UnicodeUtils
```
tomecho: Didn't know that was a thing! I tried ruby's built in and it couldnt handle unicode.
image

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -48,17 +47,65 @@ function costCalculation(){
...
-        alert('failed to retreive rental cost');
-      }
+      error: function () { alert('failed to retreive rental cost'); }
     });
    }
 }
 
+
+function userResultHandleClick(e) {
+  var userId = $(this).attr("id").substr(10); //select the user id
```
tomecho: the id's are in the format `userResult#` the `#` being the id.

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -48,17 +47,65 @@ function costCalculation(){
...
+      $(".userResult").click(userResultHandleClick);
+
+      //need to reformat links
+      $("#userResults .pagination a").each(modalSearchResultsFormatPaginationLinks);
+    },
+    error: function () { alert("Failed to retreive users."); }
+  });
+}
+
+function getQueryString( field, url ) {
```
tomecho: I copied that source off of an existing library.

app/assets/javascripts/rentals/new.js.erb
dfaulken: Okay, seems like a regex would be clearer and less prone to failure then.

app/assets/javascripts/rentals/new.js.erb
dfaulken: In that case, it belongs in `vendor`, and its source should be credited.

But I think URI.js is nicer than this, and if you include it properly as a library, like with rails-assets, then you'll get updates over time as the maintainers of that library make patches and bugfixes.

Whereas in its current form, that load is on us.

app/models/user.rb
dfaulken: I see, so it's not about it 'being Unicode' so much as this being an app internationalized in English.

spec/controllers/home_controller_spec.rb
tomecho: yes, but they were starting at 1111 instead of just 1 thinking there would be other rentals created by `create :mock_rental` which has a `sequence :reservation_id`, therefore to avoid collisions  they started there.

db/seeds.rb
tomecho: It is a permission to test if a user is able to change the value of a field.

app/controllers/rentals_controller.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'will_paginate/array'
```
tomecho: I think I was debugging and thought that was necessary, it works perfectly fine when i remove it.

spec/controllers/home_controller_spec.rb
dfaulken: Well you said it yourself, there's a sequence. Why do you need to specify the ID yourself then?

db/seeds.rb
dfaulken: So you have a controller action that's completely unroutable but that serves as a placeholder for whether or not someone has the permission to do something on a different page?

spec/controllers/home_controller_spec.rb
tomecho: I changed it already. It was originally like that because when the reservation_id is set the model will not create a real reservation with aggressive-epsilon.

dfaulken: OK.

db/seeds.rb
tomecho: no there is no method in rentals_controller.rb called cost_adjustment.

dfaulken: OK.

PR #218, opened 2016-12-14
Documents model

### LINKED ISSUE #182: Finish 'documents' model
# Finish implementing 'documents' model

Looks like the current 'documents' model is only a placeholder, as it only has a filename column
- 'documents' needs to be finished to actually store a file in the database
- Make new documents model polymorphic (see 'notes' table for reference)
### Add documents to incurred incidentals
- Remove the 'incurred_incidentals_documents' table
- Use new polymorphic 'documents' instead

tomecho:
fixes #182


app/assets/javascripts/incurred_incidentals/form.js
```diff
@@ -0,0 +1,27 @@
+$(document).ready(function () {
+  $("#multiUploadMoreFields").click(function () {
+    if($(".multi-upload-group:visible").length < 1) {
```
dfaulken: `== 0`, I think is more communicative here

app/assets/javascripts/incurred_incidentals/form.js
```diff
@@ -0,0 +1,27 @@
...
+      fileInput.attr("name",fileInput.attr("name").replace(/\d/,num)); //change the number in the id
+
+      var descInput = group.find('.multi-upload-desc'); //find file desc in the group
+      descInput.attr("name",descInput.attr("name").replace(/\d/,num)); //change the number in the id
+    }
+  });
+
+  $(".multiUploadDeleteField").click(function (e) {
+    if($(".multi-upload-group").length > 1) {
+      e.currentTarget.parentElement.remove();
```
dfaulken: A little weird to be in jQuery, but to use the native Javascript DOM descriptors.

Should be `$(this).parent().remove();`.

app/assets/stylesheets/incurred_incidentals/form.css
```diff
@@ -0,0 +1,13 @@
+.multi-upload-desc {
+  width: inherit !important;
```
dfaulken: Are we sure that these `!important`s are necessary?

I ask because it's pretty easy to get into an arms race with those.

app/controllers/documents_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  before_action :set_document
+
+  def show
+    send_data @document.fetch_file, filename: @document.original_filename
+  end
+
+  private
+
+  def set_document
+    @document = Document.find params[:id]
```
dfaulken: Should be `params.require(:id)`, right?

app/controllers/incurred_incidentals_controller.rb
```diff
@@ -56,6 +57,20 @@ def update
 
   private
 
+  def upload_documents
+    # only do this on sucess and with a file
+    if @incurred_incidental.errors.empty? && params[:file]
+      # only allow types of uploaded file
+      params.require(:file).transform_values! { |uf| uf if uf.class == ActionDispatch::Http::UploadedFile }
```
dfaulken: I like `is_a?` better than `class == `

app/controllers/incurred_incidentals_controller.rb
```diff
@@ -56,6 +57,20 @@ def update
...
+
+      params.require(:file).to_unsafe_h.map(&:itself).each do |id, uploaded_file|
```
dfaulken: So what's this do?

app/models/document.rb
```diff
@@ -1,8 +1,42 @@
...
-    "#{Rails.root}/storage/#{Rails.env}/#{filename}"
+  attr_readonly :filename # this will be set in the before action
+  attr_accessor :uploaded_file # dummy temp field to hold file
+
+  def fetch_file_as_base64
+    Base64.encode64(fetch_file) if persisted?
+  end
+
+  def fetch_file
+    File.read(Rails.root.join('storage', Rails.env.to_s, filename)) if persisted?
```
dfaulken: Seems like a helper method for the filepath might be helpful.

app/models/document.rb
```diff
@@ -1,8 +1,42 @@
...
+                        :picture
+                      else
+                        :other
+                      end
+
+      File.open(Rails.root.join('storage', Rails.env.to_s, filename), 'wb') do |file|
+        file.write(uploaded_file.read)
+      end
+    rescue StandardError
+      raise StandardError, 'Failed to properly write file'
```
dfaulken: A little odd to swallow specific exceptions in order to raise a generic one. Does this get caught somewhere?

app/models/rental.rb
```diff
@@ -140,10 +140,16 @@ def self.to_json_reservations
     arr
   end
 
-  def sum_amount
-    due = financial_transactions.where.not(transactable_type: Payment.name).sum(:amount)
-    paid = financial_transactions.where(transactable_type: Payment.name).sum(:amount)
-    due - paid # costs - payments
+  def sum_charges
```
dfaulken: Nitpick, but `sum_charges` sounds like an imperative, where this method is more of a query than a command. I might go with `charge_sum`.

app/views/incurred_incidentals/show.html.erb
```diff
@@ -61,25 +61,41 @@
   </div>
 </div>
 
+<% if !@incurred_incidental.try(:documents).empty? %>
```
dfaulken: That's the same as `present?` right?

app/views/incurred_incidentals/show.html.erb
```diff
@@ -61,25 +61,41 @@
...
+        </div>
+        <div class="panel-body no-gutter" style="max-height: 10;overflow-y: scroll;">
+          <% @incurred_incidental.documents.each do |document| %>
+            <div class="row">
+              <div class="col-xs-3">
+                <%= document.description %>
+              </div>
+              <div class="col-xs-9 pull-right">
+                <% if document.picture? %>
+                  <a href=<%= document_path(document) %>>
```
dfaulken: Any reason not to use `link_to` here also?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.69
+    "covered_percent": 69.44
```
dfaulken: Whoops

spec/controllers/rentals_controller_spec.rb
```diff
@@ -266,7 +266,7 @@
       one = create :user, email: 'billy@example.com' # name will not be related to billy
       two = create :user, first_name: 'billy' # email will not be related to billy
       get :search_users, params: { user_search_query: 'billy' }
-      expect(assigns.fetch :users).to contain_exactly one, two # will be exact match
+      expect(assigns.fetch(:users)).to contain_exactly one, two # will be exact match
```
dfaulken: Is this a Rubocop thing?

spec/models/document_spec.rb
```diff
@@ -2,17 +2,55 @@
...
+      expect(document.filename).to be_present
+    end
+
+    context 'sets filetype' do
+      it 'sets a filetype' do
+        expect(document.filetype).to be_present
+      end
+
+      it 'sets picture' do
+        expect(document).to be_picture # based on the fixture in spec/fixtures/file.png
```
dfaulken: Nice! This is a wonderful use of the dynamic predicate matcher.

spec/models/rental_spec.rb
```diff
@@ -97,14 +97,21 @@
...
       sum_amount = FinancialTransaction.where(rental: @rental).map(&:amount).inject(:+)
-      expect(@rental.sum_amount).to eq(sum_amount)
+      expect(@rental.balance).to eq(sum_amount)
+    end
+
+    it 'returns the cost-payments' do
+      sum_amount = @rental.financial_transactions.where.not(transactable_type: Payment.name).sum(:amount)
+      create(:financial_transaction, transactable: create(:payment), amount: sum_amount, rental: @rental)
+
+      expect(@rental.balance).to be_zero # fully payed
```
dfaulken: typo

app/assets/stylesheets/incurred_incidentals/form.css
tomecho: yes they are important, i dont remember why exactly but it will be the wrong width if i dont.

app/controllers/incurred_incidentals_controller.rb
tomecho: ohhhh i like that too.

app/assets/stylesheets/incurred_incidentals/form.css
dfaulken: 👍

app/controllers/incurred_incidentals_controller.rb
tomecho: Line 64 strips out anything except anything but `ActionDispatch::Http::UploadedFile` types.
Line 66 iterates over the params, I make it an unsafe hash so I allow everything and so I have access to the map function.  (map is depricated in ActionParameters.)

dfaulken: So why do you need to do that?

app/assets/javascripts/incurred_incidentals/form.js
```diff
@@ -1,6 +1,6 @@
 $(document).ready(function () {
   $("#multiUploadMoreFields").click(function () {
-    if($(".multi-upload-group:visible").length < 1) {
+    if($(".multi-upload-group:visible").length == 1) {
```
dfaulken: Whoops, think you meant 0.

app/controllers/documents_controller.rb
tomecho: right right, i dont know why but sometimes i just forget about strong parameters.

app/models/document.rb
tomecho: No it does not get caught, I wanted to rewrite the exception message.

dfaulken: For what purpose? Like I said, doing this glosses over whatever actually went wrong.

spec/controllers/rentals_controller_spec.rb
tomecho: yes

dfaulken: OK.

app/controllers/incurred_incidentals_controller.rb
tomecho: i don't remember finding one and I agree, I would rather find a safer way to do it than simply allowing everything.   This line here does something super simple, it takes a hash like  `{ "1" => UploadedFile, "2" => UploadedFile }` and transforms it into `[["1", UploadedFile],["2", UploadedFile]]` then the each block takes the parts uses them to store the file.

dfaulken: How about `params.require(:file).permit!` followed by

```ruby
params[:file].each_pair do |id, file|
  # ...
end

Comment (tomecho) [inline app/controllers/incurred_incidentals_controller.rb]: Ok I still want to do this part though
image

Comment (tomecho) [inline app/controllers/incurred_incidentals_controller.rb]: to be certain I am recieving just uploaded files.

Comment (dfaulken) [inline app/controllers/incurred_incidentals_controller.rb]: It seems simpler to me to do that within the iteration.

```ruby
params.file.each_pair do |id, file|
  next unless file.is_a? ActionDispatch::Http::UploadedFile
  #...
end
```

PR #220, opened 2017-01-04
Damage tracking

### LINKED ISSUE #111: Add Damage tracking
We would like the ability to track damage per vehicle.

Idea: When an Incidental occurs (based on type) create a record of damage on the cart. Theoretical models would include:

Damage Type:
- ~~Name~~ inherits from incurred_incidental

Damage:
- ~~Cart Id~~ inherits from incurred_incidental
- ~~Damage Type~~ will be too dynamic, should use text field
- ~~Rental Id?~~ inherits from incurred_incidental
- Estimated Cost to repair
- Actual Cost to repair
- ~~Reservation used to make the repair (time used by a mechanic)~~ hold
- Date of Damage
- Date of Repair
- User who made repair
- ~~Description (text)~~ use this instead of damage type
- Location of incident
- ~~Comments (non public?),~~ can take place on incidental

May live in the API, or may live on the app.

tomecho:
fixes #111 
be aware of #221


app/controllers/damages_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    else
+      flash[:warning] = 'Error updating Damage'
+      render 'edit'
+    end
+  end
+
+  private
+
+  def set_damage
+    @damage = Damage.find(params[:id])
```
owlee: find_damage instead of set?

app/controllers/holds_controller.rb
```diff
@@ -55,8 +55,17 @@ def set_hold
     @hold = Hold.find(params[:id])
   end
 
+  def new_hold_params
+    # find_by returns 1 or nil
+    { damage: Damage.find_by(id: params.fetch(:damage_id, nil)), item_id: params.fetch(:item_id, nil) }
```
owlee: ?

db/db_yml/incidental_types.yml
```diff
@@ -1,15 +1,19 @@
 - name: 'Late'
   description: 'Cart was checked in late'
   base: 10
+  damage_tracked: false
```
owlee: ?

owlee: resolved!

app/controllers/holds_controller.rb
owlee: resolved

PR #231, opened 2017-01-27
changed spire_id to string

tomecho:


spec/controllers/application_controller_spec.rb
```diff
@@ -132,7 +132,7 @@ def test_no_permission; end
 
       it 'raises an error if no user with shibboleth id exists' do
         user = create(:user)
-        request.env['fcIdNumber'] = (user.spire_id + 1).to_s
```
josephkelley: The change made here results in different behavior. Maybe look into what this test is for if it still passes? : )

PR #236, opened 2017-01-31
update renter references from rental.user to rental.renter

josephkelley:
There were a few lingering references to rental.user in the app after #213, causing an error on rental processing pages and elsewhere. This updates them from rental.user->rental.renter. Please double check that none of them were supposed to refer to the creator of the rental.

### DIFF
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -10,7 +10,7 @@ class Hold < ActiveRecord::Base
   validates :start_time, :end_time, date: { after: Date.current, message: 'must be after current date. ' }
   validates :end_time, date: { after: :start_time, message: 'must be after start time' }
 
-  def check_conflicting_rentals
+  def handle_conflicting_rentals
     conflicting_rentals = Rental.where('start_time >= :hold_start_time AND end_time <= :hold_end_time',
                                        hold_start_time: start_time,
                                        hold_end_time: end_time)
@@ -23,7 +23,7 @@ class Hold < ActiveRecord::Base
                             renter_id: curr_rental.renter_id, department_id: curr_rental.department_id,
                             start_time: curr_rental.start_time, end_time: curr_rental.end_time)
     new_rental.create_reservation
-    new_rental.save
+    new_rental.save!
     curr_rental.cancel!
     ReplacementMailer.replacement_email(curr_rental.renter, self, curr_rental, new_rental).deliver_now
   rescue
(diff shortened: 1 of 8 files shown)


app/models/hold.rb
```diff
@@ -20,15 +20,15 @@ def check_conflicting_rentals
 
   def replace_rental(curr_rental)
     new_rental = Rental.create(item_type_id: curr_rental.item_type_id,
-                               user_id: curr_rental.user_id, department_id: curr_rental.department_id,
```
tomecho: how did a test not catch this error?

josephkelley:
I added the tests you asked for and found/fixed a couple issues. Funny how that happens : )

app/models/hold.rb
```diff
@@ -19,16 +19,16 @@ def check_conflicting_rentals
...
-                               start_time: curr_rental.start_time, end_time: curr_rental.end_time)
-
-    curr_rental.cancel
-    curr_rental.save
-    ReplacementMailer.replacement_email(curr_rental.user, self, curr_rental, new_rental).deliver_now
+    new_rental = Rental.new(item_type_id: curr_rental.item_type_id, creator_id: curr_rental.creator_id,
+                            renter_id: curr_rental.renter_id, department_id: curr_rental.department_id,
+                            start_time: curr_rental.start_time, end_time: curr_rental.end_time)
+    new_rental.create_reservation
+    new_rental.save
```
tomecho: what if this fails? on `new_rental.save` failure it will simply return `false` right?

spec/models/hold_spec.rb
```diff
@@ -39,4 +39,37 @@
...
   end
+
+  describe 'check_conflicting_rentals' do
+    let(:hold) { create(:hold) }
+    let(:conflicting_rental) { create(:hold_conflicting_rental) }
+    let(:future_rental) { create(:far_future_rental) }
+
+    it 'should cancel a conflicting rental' do
+      conflicting_rental
+      hold.check_conflicting_rentals
```
tomecho: does `hold.check_conflicting_rentals` replace the rental? It seems a little counter intuitive that a method named that would have a destructive action. Maybe rename it to `handle_conflicting_rentals`

spec/models/hold_spec.rb
```diff
@@ -39,4 +39,37 @@
...
+      future_rental
+      expect do
+        hold.check_conflicting_rentals
+      end.not_to change(future_rental, :rental_status)
+    end
+
+    it 'should create a new rental when there is a conflicting rental' do
+      conflicting_rental
+      expect do
+        hold.check_conflicting_rentals
```
tomecho: Might also want to do an assertion for sending an email

     expect do
        some thing that sends an email
     end.to change{ActionMailer::Base.delivery_count}.by(1)

Issue #232, opened 2017-01-30
incurred_incidental#show format data in user friendly way


tomecho:
![image](https://cloud.githubusercontent.com/assets/2114323/22436142/e828b740-e6f0-11e6-97d0-49779e77e639.png)
- Show amount with dollar sign hint use `number_to_currency`
- format dates

josephkelley:
We also have date/currency beautification work to do in:
- the financial transaction table
- incurred_incidentals#index
- payment tracking

How would we like to format dates? In the screenshot above, we have two different stylings (Mon dd,time and yyyy-mm-dd). I get why that's the case above, seems reasonable in context, but there are other places in the app where we use mm-dd-yyyy and mm/dd/yyyy, and it would probably be best to narrow that down to one or two variations in formatting.

Maybe "Mon dd, yyyy time" for notes like above (or whenever time relevant?) and "mm-dd-yyyy" for all other dates?

dfaulken:
> Maybe "Mon dd, yyyy time" for notes like above (or whenever time relevant?) and "mm-dd-yyyy" for all other dates?

I can't think of any reason to represent a date differently when it's paired with a time.

josephkelley:
I just liked the way it looked in the Notes box in @tomecho's screenshot above, but for simplicity's sake I'll make it mm-dd-yyyy everywhere.

dfaulken:
http://ux.stackexchange.com/questions/16095/how-to-display-date-and-time

PR #237, opened 2017-01-31
Simplify date formatting, format currency

josephkelley:
#232. Uses default date formatting in places where we either didn't format at all or formatted a little strangely, and adds currency formatting where it was missing.


josephkelley:
Good call, I should have questioned the status quo. I'll remove each `try` in the lines I changed, especially since they don't do anything where `to_date` was called.

dfaulken:
It's also worth noting that in general chaining any method other than `try` on `try` isn't a good idea, since the whole point of it is that it passes `nil` through.

PR #245, opened 2017-02-08
Editing documents

### LINKED ISSUE #234: incurred_incidental#edit cannot edit documents
does not populate documents fields therefore there is no way to remove them from the application.

Should research the incurred_incidental model `accepts_nested_attributes_for :documents`

### LINKED ISSUE #243: rentals can be invalid after creation
image
- solution: only verify if not persisted

tomecho:
fixes #234


spec/controllers/holds_controller_spec.rb
```diff
@@ -46,23 +46,23 @@
...
-        #Timecop.travel Time.current+1.day
-        #expect(hold.start_time).to be < Time.current
-        #post :lift, params: { id: hold.id }
-        #hold.reload
-        #expect(hold).to be_valid
-        #expect(hold.active).to be false
-      #end
-    #end
-  #end
+  # describe 'POST #lift' do
```
josephkelley: Let's delete this :)

PR #257, opened 2017-03-02
dangermouse = 1

tomecho:


sherson:
dangermouse!

tomecho:
@dfaulken 
i know by default net:http does not verify certs,
however http does verify certs is it safe to disable cert verification?

Issue #274, opened 2017-03-09
relevant errors


tomecho:
controllers should be less ambiguous when failing
return the errors not a general message
i see this in 
- [X] holds 
- [X] rentals
- [ ] others

tomecho:
![image](https://cloud.githubusercontent.com/assets/2114323/23764056/1de42332-04ca-11e7-988f-a0dcc8e30285.png)
should always do this

tomecho:
closing this issue, will open a different one if something specific arises

PR #276, opened 2017-03-09
better error messages on rental creation failure

tomecho:
part of #274


tomecho:
no the api still does the same thing, we had added errors to the instance of `Rental` but our error message never showed them.

Issue #263, opened 2017-03-08
Canceled rentals still show a balance due


tomecho:
Some rentals should have their balance adjusted to zero when canceled others should keep the balance (depending on cancellation time)

tomecho:
closed by #282

PR #282, opened 2017-03-10
Cancel Charge fees

owlee:
#263 

Changing :amount to :initial_amount, and changing every call to amount to be value()


app/models/financial_transaction.rb
```diff
@@ -1,19 +1,22 @@
...
+  after_save :send_updated_invoice
+
   belongs_to :rental
   belongs_to :transactable, polymorphic: true
 
-  after_save :send_updated_invoice
-
   validates :adjustment, :rental_id, :amount, presence: true
-  validates :amount, numericality: { greater_than: 0 }
+  validates :amount, numericality: { greater_than: 0 }, presence: true
```
tomecho: why validate presence twice?

owlee: :man_cartwheeling: 💣 💥

app/models/rental.rb
```diff
@@ -143,11 +146,11 @@ def self.to_json_reservations
   end
 
   def sum_charges
-    financial_transactions.where.not(transactable_type: Payment.name).sum(:amount)
+    financial_transactions.where.not(transactable_type: Payment.name).inject(0) {|total, transaction| total + transaction.balance}
```
owlee: Need to double check if works. Simpler way would be a map to balance

owlee:
@tomecho Should I change :amount from incurred_incidental to :initial_amount also? It seems pretty harmless w/o the ':adjustments' that financial_transaction has.

nawilliams:
Needed to change .amount to .initial_amount in the views too. 
when creating rental

I think I caught them all here: 0718612d1a62123302523b459c06f69bc677482f

PR #284, opened 2017-03-10
rubocop auto corrected these

tomecho:


tomecho:
actually no this is silly, we wont do this

PR #294, opened 2017-03-15
Date ranges fix (pricing related)

### LINKED ISSUE #291: Calculating date ranges is incorrect
image

- Number of days is incorrect, in the form
- as well as in the financial transaction calculation

We should be counting starting on the same day as ending as one full day.  Basically add 1 to whatever we are calculating now.

tomecho:
fixes #291


tomecho:
needs to be discussed at a code review, this changes pricing.

PR #298, opened 2017-03-25
fixed redirecting current window

### LINKED ISSUE #270: View All Rentals
Clicking on any rental should only open up a new tab with the rental#show. But, it opens up a new tab with the show AND opens up the show on View All.

nawilliams:
Fixes #270 as per [documentation](https://fullcalendar.io/docs/mouse/eventClick/).

### DIFF
--- a/app/assets/javascripts/rentals/index.js
+++ b/app/assets/javascripts/rentals/index.js
@@ -6,6 +6,7 @@ $(document).ready(function() {
 		eventClick: function(calEvent, jsEvent, view) {
 			var win = window.open(calEvent.url, '_blank');
 			win.focus();
+			return false;
 		}
 	})
 });


app/assets/javascripts/rentals/index.js
```diff
@@ -6,6 +6,7 @@ $(document).ready(function() {
 		eventClick: function(calEvent, jsEvent, view) {
 			var win = window.open(calEvent.url, '_blank');
 			win.focus();
+			return false;
```
owlee: What does returning false do (functionally) in this context?

nawilliams: From the Documentation I linked:
"Normally, if the Event Object has its url property set, a click on the event will cause the browser to visit the event's url (in the same window/tab). Returning false from within your function will prevent this from happening."

It looks like by returning false it breaks out of the event, stopping the code from executing that opens the url in the current window.

PR #299, opened 2017-03-25
Invoice remove underscore between picked and up

### LINKED ISSUE #267: remove underscore between picked up
image

nawilliams:
I'm dumb and was looking at the rentals page not the invoice page. Now actually fixes #267.
@tomecho If you think it's more aesthetic in uppercase we can change it, we just need to make sure it's consistent between the rentals and invoice pages.


owlee:
Remember to fix Travis errors, update your branch from master, and maybe include screenshots in future.

PR #313, opened 2017-04-07
Holds on in progress

### LINKED ISSUE #268: holds on items that are in progress do not work
should notify user when creating a hold on an item that is already picked up

tomecho:
fixes #268 
and an error we never thought of where holds consider any rental between the date ranges to be a conflict


tomecho:
image
Example of what happens when an in progress item is put on hold

app/controllers/holds_controller.rb
```diff
@@ -18,6 +19,12 @@ def create
     if @hold.save
       flash[:success] = 'Hold Successfully Created'
       @hold.handle_conflicting_rentals
+
+      # rental that has already started and conflicts with hold, needs to be handled
+      # by a csr
+      if rental = @hold.conflicting_ongoing_rental
+        flash_message :warning, "Ongoing rental with item #{url_for rental}".html_safe
```
josephkelley: Let's make the URL here more user friendly. Maybe make "item" clickable with a link to the `url_for`?

app/controllers/holds_controller.rb
```diff
@@ -31,6 +38,12 @@ def update
     if @hold.update(hold_params)
       flash[:success] = 'Hold Successfully Updated'
       @hold.handle_conflicting_rentals
+
+      # rental that has already started and conflicts with hold, needs to be handled
+      # by a csr
+      if rental = @hold.conflicting_ongoing_rental
+        flash_message :warning, "Ongoing rental with item #{url_for rental}".html_safe
```
josephkelley: See above.

tomecho:
@josephkelley plz rereview

PR #333, opened 2017-04-20
removed all fields, they can just look at financial transactions, whi…

### LINKED ISSUE #317: Changing invoices
* clean up the invoice information in general
    * it probably shouldn’t say “amount charged” on the invoice page, rename that to something else we're not actually charging them
    * remove reference to pro-card online on the invoice page 

invoice is a mailing template and also a regular view

tomecho:
…ch totals amount due at bottom anyway

fixes #317 

`rental#show`
image

printed invoice
[ProbableEngine.pdf](https://github.com/umts/probable-engine/files/944447/ProbableEngine.pdf)

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class Payment < ApplicationRecord
   has_one :financial_transaction, as: :transactable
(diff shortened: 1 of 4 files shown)


tomecho:
:+1: thanks @josephkelley

PR #340, opened 2017-04-25
Rental reservation event

### LINKED ISSUE #308: move rental.create_reservation to before save event
image
Doing it like this can cause an error when one is successful but the other fails

tomecho:
fixes #308

### DIFF
--- a/app/models/rental.rb
+++ b/app/models/rental.rb
@@ -9,7 +9,8 @@ class Rental < ActiveRecord::Base
   has_one :financial_transaction, as: :transactable
 
   after_create :create_financial_transaction
-  before_save :create_reservation
+  # create reservation unless it has already been created
+  before_save :create_reservation, unless: proc { |rental| rental.reservation_id.present? }
 
   belongs_to :creator, class_name: User
   belongs_to :renter, class_name: User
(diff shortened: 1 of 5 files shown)


app/models/rental.rb
```diff
@@ -88,14 +90,16 @@ def renter_is_assignable
...
     begin
       reservation = Inventory.create_reservation(item_type.name, start_time, end_time)
+      raise 'Reservation UUID was not present in response.' unless reservation[:uuid].present?
+
       self.reservation_id = reservation[:uuid]
       self.item = Item.find_by(name: reservation[:item][:name])
     rescue => error
       errors.add :base, error.inspect
-      return false
+      throw :abort
```
tomecho: welcome to the rails5 way of aborting `before_*`

app/models/rental.rb
```diff
@@ -167,14 +171,6 @@ def duration
     (end_time.to_date - start_time.to_date).to_i + 1
   end
 
-  # this method seems really inefficient
-  # def self.with_balance_due
-  # Rental.select { |x| x.balance > 0 }
-  # end
-
-  # private
-  attr_accessor :skip_reservation_validation
```
tomecho: unless i am mistaken this isnt used

spec/models/rental_spec.rb
```diff
@@ -345,22 +352,6 @@
     end
   end
 
-  describe '#create_reservation' do
```
tomecho: shouldn't call this method directly, it is a before_save event now.

PR #341, opened 2017-04-27
Add ref to payment

### LINKED ISSUE #323: Reference field on financial_transaction#new (which is new payment)
The reference number will be batch number for recharge submit to a&f

tomecho:
closes #323

### DIFF
--- a/app/views/payment_tracking/index.html.erb
+++ b/app/views/payment_tracking/index.html.erb
@@ -36,6 +36,7 @@
         <th>Renter Department/Name</th>
         <th>Days Since Reservation</th>
         <th>Balance Due</th>
+        <th>Payment Reference</th>
         <th class="min-width nosortcol">Actions</th>
       </tr>
     </thead>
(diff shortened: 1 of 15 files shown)


app/views/payment_tracking/index.html.erb
```diff
@@ -48,6 +49,7 @@
           <td><%= if rental.renter.department then rental.renter.department.name else rental.renter.full_name end %></td>
           <td><%= ((Time.current-rental.created_at) / 1.days ).floor %></td>
           <td><%= number_to_currency rental.balance %></td>
+          <td><%= rental.payments.map { |ft| ft.transactable.reference }.join ', ' %></td>
```
tomecho: This is safe if there are no payments, it will return a empty `ActiveRecord::Relation`, performing a map on that object will yield an empty array joined by , is just nil.

owlee: Is it worth checking for payments.empty? beforehand? Stylistic choice...

rent.pay.map {} unless pay.empty?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.6
+    "covered_percent": 98.58
```
owlee: tsk tsk tsk @tomecho

Issue #321, opened 2017-04-14
attach recharge form to invoice


tomecho:
- [ ] and the printable one
- [ ] the mailed one

tomecho:
maybe add image to printable web page one
![image](https://cloud.githubusercontent.com/assets/2114323/25546144/c2b159b6-2c2f-11e7-9067-a1ef4fc6785b.png)

PR #342, opened 2017-04-28
Add recharg to invoice

tomecho:
partial fix to #321


app/views/rentals/_transaction_detail.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="table table-sm">
+<table class="table table-striped table-bordered table-hover">
```
tomecho: this is totally unrelated but was pissing me off

tomecho: this table is next to the table on `rental#show` and should match

PR #357, opened 2017-10-04
Integration testing

enragednuke:
Probably won't work on Travis first pass. Likely going to end up doing a bunch of extra commits to fix travis and the like.

### DIFF
--- a/app/models/inventory_mock.rb
+++ b/app/models/inventory_mock.rb
@@ -59,11 +59,15 @@ class InventoryMock
   end
 
   def self.create_reservation(item_type, start_time, end_time)
-    JSON.parse("{\"id\": \"#{SecureRandom.uuid}\",
-    \"start_time\": \"#{start_time}\",
-    \"end_time\": \"#{end_time}\",
-    \"item_type\": \"#{item_type}\",
-    \"item\": \"Dummy Data\"}")
+    {
+      uuid: SecureRandom.uuid,
+      start_time: start_time,
+      end_time: end_time,
+      item_type: item_type,
+      item: {
+        name: "Fiore"
+      }
+    }
   end
 
   # maybe in the future add a few helper methods like update_start_time(uuid,start_time)
--- a/app/models/rental.rb
+++ b/app/models/rental.rb
@@ -95,12 +95,21 @@ class Rental < ActiveRecord::Base
   def create_reservation
     throw :abort unless valid? # check if the current rental object is valid or not
     begin
-      reservation = Inventory.create_reservation(item_type.name, start_time, end_time)
+      if Rails.env.development?
+        reservation = InventoryMock.create_reservation(item_type.name, start_time, end_time)
+      else
+        reservation = Inventory.create_reservation(item_type.name, start_time, end_time)
+      end
+
       raise 'Reservation UUID was not present in response.' unless reservation[:uuid].present?
 
       self.reservation_id = reservation[:uuid]
       self.item = Item.find_by(name: reservation[:item][:name])
     rescue => error
+      # vague begin-rescue blocks like this make it very hard to determine
+      # the error causes during development
+      raise error if Rails.env.development?
+
       errors.add :base, error.inspect
       throw :abort
     end
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -11,10 +11,10 @@
 # if you're sharing your code publicly.
 
 development:
-  secret_key_base: 8f85b60688ca164debb7960ac838f074dbb54def7299ee7f2363afe5da783d5595b741dd2c2c17d1ee5f3bc6de18dc540240fc8064525ba1c157ce604143c8db
+  secret_key_base: dc889de5ddd86a8f57ec040472800e3703c425d70eaf3721cbdc7e759e9b1125abb6bb45c073dd1c3045ee036c2201a37803804085308c4c6671e11ebbe6a9a5
 
 test:
-  secret_key_base: ce60ba18c3631123069bd459b0b2e755effdf5be8017786760cc063046b3bbf318515ed5fa1204aab7a98be14c3a6b9ef03aaa9c03658648fe6818de7d1e9281
+  secret_key_base: dc889de5ddd86a8f57ec040472800e3703c425d70eaf3721cbdc7e759e9b1125abb6bb45c073dd1c3045ee036c2201a37803804085308c4c6671e11ebbe6a9a5
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.64
+    "covered_percent": 56.43
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 56.43
+    "covered_percent": 98.59
   }
 }
--- a/spec/features/rental_creation_spec.rb
+++ b/spec/features/rental_creation_spec.rb
@@ -2,6 +2,7 @@ require 'rails_helper'
 
 feature 'Creating a new Rental' do
 
+  # need to set selenium driver at some point to test javascript related features
   # before(:all) do
   #   Capybara.current_driver = :selenium
   # end
@@ -12,7 +13,8 @@ feature 'Creating a new Rental' do
 
   before(:each) do
     @a_user = create(:admin_user)
-    allow_any_instance_of(ApplicationController).to receive(:current_user).and_wrap_original { @current_user = @a_user }
+    # this line is broken, need to find some way to set @current_user in rentals_controller
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(@a_user)
 
     @item_type = create(:item_type)
 
@@ -22,7 +24,8 @@ feature 'Creating a new Rental' do
 
   scenario 'Form fields are autopopulated' do
     visit new_rental_url
-    # binding.pry
+    # some fields are autopopulated but it's difficult to test their values
+    # might be worth to investigate in the future
     within 'form.new_rental' do
       @today = Date.today.strftime('%Y-%m-%d')
       expect(find_by_id('rental_item_type_id').value).to eql @item_type.id.to_s
@@ -32,6 +35,8 @@ feature 'Creating a new Rental' do
     end
   end
 
+  # none of the following tests work because the selenium-webdriver doesn't seem
+  # to be able to open the rental creation form. investigate in future
   @javascript
   scenario 'Creating a Rental with valid parameters' do
     visit new_rental_url
--- a/spec/features/rental_creation_spec.rb
+++ b/spec/features/rental_creation_spec.rb
@@ -1,94 +1,74 @@
 require 'rails_helper'
 
-feature 'Creating a new Rental' do
-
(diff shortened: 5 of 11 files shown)


tomecho:
Something tells me travis wont allow chrome web driver to work on their system, we might want to skip integration tests on travis

enragednuke:
It is possible to use the firefox driver, I just didn't have Firefox installed in my local computer. 

Either way, it would be convenient to have travis skip integration tests since it has to run a server too and integration tests a long time as is.

.travis.yml
```diff
@@ -26,7 +26,6 @@ before_script:
   # test to see if the db is an issue
   - mysql -e 'CREATE DATABASE probabledev;'
   - mysql -e 'CREATE DATABASE probabletest;'
-  - mysql -e "SET PASSWORD FOR 'travis' = PASSWORD('probable');"
   - mysql -e "GRANT ALL PRIVILEGES on probabledev.* TO 'travis';"
```
dfaulken: Do you need a development database to run integration tests?

enragednuke: Yes, the feature tests run off of the development database

app/models/rental.rb
```diff
@@ -95,7 +95,14 @@ def renter_is_assignable
   def create_reservation
     throw :abort unless valid? # check if the current rental object is valid or not
     begin
-      reservation = Inventory.create_reservation(item_type.name, start_time, end_time)
+
+      # Because integration testing using development environment (not fixable
```
Anbranin: Hmm. Do you mean that the `RAILS_ENV == 'development'`? That doesn't seem right.

app/models/rental.rb
```diff
@@ -207,4 +214,14 @@ def create_financial_transaction
     rental_amount = Rental.cost(start_time.to_date, end_time.to_date, item_type)
     FinancialTransaction.create rental: self, amount: rental_amount, transactable_type: self.class, transactable_id: id
   end
+
+  private
+
+  def reservation_response(item_type, start_time, end_time)
+    if Rails.env.development?
+      InventoryMock.create_reservation(item_type.name, start_time, end_time)
```
Anbranin: Clever fix, in any case.

app/controllers/application_controller.rb
```diff
@@ -14,7 +14,7 @@ class ApplicationController < ActionController::Base
   # Set auto papertrail
   before_action :set_paper_trail_whodunnit
 
-  rescue_from RuntimeError, Exception, with: :render_500 unless Rails.env.development?
+  # rescue_from RuntimeError, Exception, with: :render_500 unless Rails.env.development?
```
Anbranin: Although I do disagree with catch-all rescuing, why just comment this out?

config/secrets.yml
```diff
@@ -11,10 +11,10 @@
...
 
 development:
-  secret_key_base: 8f85b60688ca164debb7960ac838f074dbb54def7299ee7f2363afe5da783d5595b741dd2c2c17d1ee5f3bc6de18dc540240fc8064525ba1c157ce604143c8db
+  secret_key_base: dc889de5ddd86a8f57ec040472800e3703c425d70eaf3721cbdc7e759e9b1125abb6bb45c073dd1c3045ee036c2201a37803804085308c4c6671e11ebbe6a9a5
 
 test:
-  secret_key_base: ce60ba18c3631123069bd459b0b2e755effdf5be8017786760cc063046b3bbf318515ed5fa1204aab7a98be14c3a6b9ef03aaa9c03658648fe6818de7d1e9281
+  secret_key_base: dc889de5ddd86a8f57ec040472800e3703c425d70eaf3721cbdc7e759e9b1125abb6bb45c073dd1c3045ee036c2201a37803804085308c4c6671e11ebbe6a9a5
 
 # Do not keep production secrets in the repository,
```
Anbranin: Shouldn't this file be in the gitignore?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.6
+    "covered_percent": 56.89
```
Anbranin: What happened here??

db/seeds.rb
```diff
@@ -1,5 +1,7 @@
 include InventoryExceptions
 
+dept = Department.create(name: '_test')
+User.create(first_name: 'Apples', last_name: 'Sauce', spire_id: '00000000', email: "applesauce@xyz.net", phone: '1248279421', department: dept)
```
Anbranin: xyz.net could actually be a URL, why not just use example.com for everything? :)

spec/features/rental_creation_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+          #expect(page).to have_button('rentalSubmit', disabled: false)
+          puts page.html
+
+          #click_button 'rentalSubmit'
+          page.save_screenshot('page1.png')
+          find_button('rentalSubmit').trigger('click')
+          page.save_screenshot('page2.png')
+          accept_alert
+          #page.find_by_id('rentalSubmit').click
+          #page.driver.browser.switch_to.alert.accept
```
Anbranin: I'm assuming this isn't quite done due to the amount of commented out code, but it looks super cool so far.

coverage/.last_run.json
enragednuke: Ran solely integration tests, not whole suite. Can disregard.

config/secrets.yml
enragednuke: /shrug

It wasn't there to start with so I didn't mess with it.

app/controllers/application_controller.rb
enragednuke: We ("we" being Tom and I, since we did a lot of the recent commits together) were using this via d8bdc10 to get travis to throw errors instead of just returning the 500 page (so we could see what the errors were)

app/models/rental.rb
enragednuke: Feature testing requires an actively running server, which in turn causes the tests to run off the development database. This has been documented in many things selenium-related. I assume it's not fixable because I don't remember a case that *didn't* use the development database then just use DatabaseCleaner to rollback the database after the tests.

enragednuke: For what it's worth, I have tried to manually set the environment during the tests and change it back after, which didn't work.

db/seeds.rb
enragednuke: This was just part of the debugging me and Tom did, I doubt this will stay exactly like this when it's fully fixed.

spec/features/rental_creation_spec.rb
enragednuke: Yeah, not done yet. Travis, for some reason, cannot see the submit button even though we've tried `puts page.html` right before the logic that deals with it and surprisingly, the button is in the html. We've also tried sleeping the code to see if it was a loading issue.

app/models/rental.rb
Anbranin: Ah okay, interesting! I guess in most scenarios (where their development database doesn't make any outside calls) it's fine for people.

spec/features/rental_creation_spec.rb
```diff
@@ -46,7 +46,7 @@
           fill_in('rental_end_time', with: Date.today.strftime('%Y-%m-%d'))
 
           expect(page).to have_button('rentalSubmit', disabled: true)
-          check('TOC')
+          # check('TOC')
```
Anbranin: Why'd you comment this out?

enragednuke: Trying to see if I can get something working using the new logic @tomecho wrote that it [removes the checkbox if a person's department is Parking](https://github.com/umts/probable-engine/blob/master/app/views/rentals/_form.html.erb#L71).

app/controllers/application_controller.rb
tomecho: It's redundant anyway seeing as we have exception notifier (i think) and rails (or apache or passenger i don't know which) will automatically load the 500 page anyway

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -81,6 +81,16 @@ $(document).ready(function () {
     costCalculation();
   });
 
+  $('#rentals-new-toc').change(function() {
+    var button = document.getElementById('rentalSubmit')
```
tomecho: could select using jquery if you want seeing as we already do that to get the onchange event.

app/models/inventory_mock.rb
```diff
@@ -59,11 +59,15 @@ def self.delete_item(_uuid)
...
-    \"item\": \"Dummy Data\"}")
+    {
+      uuid: SecureRandom.uuid,
+      start_time: start_time,
+      end_time: end_time,
+      item_type: item_type,
+      item: {
+        name: 'Fiore'
+      }
+    }
```
tomecho: turns out that entire class was written by me, i dunno why i couldn't remember that at the time.  also i dont know why I wrote out some json then parsed it rather than just creating a ruby hash directly.

app/assets/javascripts/rentals/new.js.erb
enragednuke: I was just being extra careful, might change it once everything starts working again.

.travis.yml
```diff
@@ -2,18 +2,39 @@ language: ruby
...
+  # - mysql -e 'CREATE DATABASE probabledev;'
+  # - mysql -e 'CREATE DATABASE probabletest;'
+  # - mysql -e "GRANT ALL PRIVILEGES on probabledev.* TO 'travis';"
+  # - mysql -e "GRANT ALL PRIVILEGES on probabletest.* TO 'travis';"
+  # --- #
   - bundle exec rake db:create
   - bundle exec rake db:migrate
   - bundle exec rake db:seed
   - export TZ="America/New_York"
+  - bundle exec rails s -p 3000 &
```
anmerkl: Does the `&` do anything here?

config/secrets.yml
anmerkl: 100% should be

spec/rails_helper.rb
```diff
@@ -7,10 +7,54 @@
...
+#       type 'brew install chromedriver' into a terminal.
+
+Capybara.register_driver :selenium_chrome do |app|
+  Capybara::Selenium::Driver.new(app, browser: :chrome)
+end
+
+Capybara.register_driver :selenium_chrome_headless do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: {
+      args: %w(headless disable-gpu no-sandbox disable-gpu window-size=1400,900})
```
anmerkl: friendless `}` at the end of the line

spec/features/rental_creation_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  describe 'rental creation' do
+    describe 'invalid' do
+      it 'redirects to creation page again' do
+        visit '/rentals/new'
+
+        within('form.new_rental') do
+          fill_in('rental_start_time', with: @tomorrow.strftime('%Y-%m-%d'))
+          fill_in('rental_end_time', with: Date.today.strftime('%Y-%m-%d'))
```
anmerkl: you already assigned `Date.today.strftime('%Y-%m-%d')` to `@today`, might as well use it

spec/features/rental_creation_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+          # check('TOC')
+          puts page.html
```
anmerkl: I assume this is leftover from some debugging? should remove this so console output isn't cluttered

PR #358, opened 2017-10-30
Rent multiple carts

### LINKED ISSUE #355: renting multiple carts at one time
need to be able to rent multiple carts all at once.
* ~~~functionality to query availability of selected types and quantities and show user on rental#new page~~~ this is #356 and can be done after
* still a single renter
* [x] single invoice with multiple rentals
   * shouldn't change much on financial transactions, there will just be multiple financial transactions of type  `rentals_item` for a single rental
* [x] rentals have many rented items
  * [x] should be able to create a rental doing `Rental.create items_types: [...]...`
* [x] create `rentals_items` table
  * [x] has reference to item and item type
  * [x] item type populated before save
  * [x] one reservation id per `rental_item` each item needs individual reservations
     * alternative could change aggressive epsilon so a single reservation id correlates to many reserved items
  * [x] item populated before rental save (reservation creation)
* [x] this will change incurred incidentals, we dont know which rented item caused the incidental
  * [ ] ~~~add item_id to incurred_incidental~~~ already exists
  * [x] when creating new incurred incidental, indicate the item

tomecho:
fixes #355 

This changed a lot of views, I am still stumbling upon errors.

## Views to fix
- [x] `app/views/rentals/index.html.erb` - can't search by item_type_id the same way as before
- [x] `app/views/rentals/show.html.erb` - show table of rentals_items

Edit: everything looks good as far as i can tell.  this would be a good time to use the integration tests, at least to render every view and hope that there are no errors.

### DIFF
--- a/app/assets/javascripts/rentals/new.js.erb
+++ b/app/assets/javascripts/rentals/new.js.erb
@@ -13,15 +13,20 @@ function days_between(sd, ed) {
   return Math.round(diff_ms/one_day) + 1;
 }
 
-function costCalculation(){
+function item_types_to_arr() {
   //collect item types (will always be at least [])
-  var item_types = $(".item-type-quantity-picker").toArray().reduce(function (acc,part) {
+  return $(".item-type-quantity-picker").toArray().reduce(function (acc,part) {
     var it_part = [];
     for(var i=0;i<part.value;i++){
-      it_part.push(part.name);
+      it_part.push($(part).attr("_name"));
     }
     return acc.concat(it_part);
   }, []);
+}
+
+function costCalculation(){
+  //collect item types (will always be at least [])
+  var item_types = item_types_to_arr();
 
   var raw_start_time = $('#rental_start_time').val();
   var raw_end_time = $('#rental_end_time').val();
@@ -74,5 +79,15 @@ $(document).ready(function () {
 
   $('input.cost-dependent.item-type-quantity-picker').change(function () {
     costCalculation();
+  });
+
+  $('form#new_rental').submit(function () {
+    //need to take all the item types and format them correctly; it should look like this
+    // { renter_id: 1, rental_items_attributes: { {item_type_id: 1}, {item_type_id: 4} }
+    // need a few inputs with name = "rental[rentals_items_attributes][0][item_type_id]"
+    // these will be hidden from user so they think those other inputs are the things that make it work
+    item_types_to_arr().forEach((it, ind) => {
+      $(this).append(`<input name='rental[rentals_items_attributes][${ind}][item_type_id]' value='${it}' hidden>`);
+    });
   });
 });
--- a/app/assets/javascripts/rentals/new.js.erb
+++ b/app/assets/javascripts/rentals/new.js.erb
@@ -14,32 +14,33 @@ function days_between(sd, ed) {
 }
 
 function costCalculation(){
-  var raw_item_type = null;
-  var raw_start_time = null;
-  var raw_end_time = null;
-  var amount = null;
-  var range = null;
+  //collect item types (will always be at least [])
+  var item_types = $(".item-type-quantity-picker").toArray().reduce(function (acc,part) {
+    var it_part = [];
+    for(var i=0;i<part.value;i++){
+      it_part.push(part.name);
+    }
+    return acc.concat(it_part);
+  }, []);
 
-  //TODO collect item types
-  raw_item_type = $('#rental_item_type_id').val();
-  raw_start_time = $('#rental_start_time').val();
-  raw_end_time = $('#rental_end_time').val();
-  amount = $('#rental_amount');
-  date_range = $('#date_range');
+  var raw_start_time = $('#rental_start_time').val();
+  var raw_end_time = $('#rental_end_time').val();
+  var amount = $('#rental_amount');
+  var date_range = $('#date_range');
 
-  if(!isEmpty(raw_item_type) && !isEmpty(raw_end_time) && !isEmpty(raw_start_time)) {
+  if(!item_types.length <= 0 && !isEmpty(raw_end_time) && !isEmpty(raw_start_time)) {
     if(new Date(raw_start_time) > new Date(raw_end_time)) {
       return; //do not perform a cost calculation
       //the end_date changed to start_date + previous range then call this method once again
     }
-    range = days_between(raw_start_time, raw_end_time);
+    var range = days_between(raw_start_time, raw_end_time);
 
     $.ajax({
       url: Routes.rentals_cost_path(),
       dataType: 'json',
-      data: { item_type: raw_item_type, start_time: raw_start_time, end_time: raw_end_time },
+      data: { item_types: item_types, start_time: raw_start_time, end_time: raw_end_time },
       success: function (data) {
-        amount.val(data);
+        amount.val(data["_total"]);
         date_range.val(range);
       },
       error: function () { alert('failed to retreive rental cost'); }
@@ -71,13 +72,7 @@ $(document).ready(function () {
     costCalculation();
   });
 
-  $('.include-item-type-in-rental').click(function (e) {
-
-    $('.cost-dependent').trigger('change'); //force a recalculation
-    e.preventDefault();
-  });
-
-  $('select.cost-dependent').on('change', function () {
+  $('input.cost-dependent.item-type-quantity-picker').change(function () {
     costCalculation();
   });
 });
--- a/app/assets/javascripts/rentals/new.js.erb
+++ b/app/assets/javascripts/rentals/new.js.erb
@@ -47,22 +47,7 @@ function costCalculation(){
    }
 }
 
-$('.selectable-item-type').click(function (e) {
-  if($(this).hasClass('active')) { //was active lets remove it
-    $(this).removeClass('active');
-    $(this).html($(this).attr('_name')); //set to base item type name
-  } else { //was inactive lets active it
-    //set item type to active
-    $(this).addClass('active');
-    //add input for ammount
-    $(this).html("<input type='number' step='1' value='1' class='item-type-quantity-picker cost-dependent'> " + $(this).attr('_name')); //set to base item type name
-  }
-
-  $('.cost-dependent').trigger('change'); //force a recalculation
-  e.preventDefault();
-});
-
-$(document).ready( function () {
+$(document).ready(function () {
   costCalculation(); //initial cost calculaton
 
   $("#rental_amount").change(function () {
@@ -84,6 +69,12 @@ $(document).ready( function () {
     }
 
     costCalculation();
+  });
+
+  $('.include-item-type-in-rental').click(function (e) {
+
+    $('.cost-dependent').trigger('change'); //force a recalculation
+    e.preventDefault();
   });
 
   $('select.cost-dependent').on('change', function () {
--- a/app/assets/javascripts/rentals/new.js.erb
+++ b/app/assets/javascripts/rentals/new.js.erb
@@ -15,6 +15,10 @@ function days_between(sd, ed) {
 
 function item_types_to_arr() {
   //collect item types (will always be at least [])
+
+  //because of the way im constructing the parameters to Rental.new I need rentals_items_attributes
+  //in a way such that multiple item types yield multiple rentals_items not just a single one.
+  //refer to this methods caller for more information
   return $(".item-type-quantity-picker").toArray().reduce(function (acc,part) {
     var it_part = [];
     for(var i=0;i<part.value;i++){
--- a/app/assets/javascripts/rentals/new.js.erb
+++ b/app/assets/javascripts/rentals/new.js.erb
@@ -20,6 +20,7 @@ function costCalculation(){
   var amount = null;
   var range = null;
 
+  //TODO collect item types
   raw_item_type = $('#rental_item_type_id').val();
   raw_start_time = $('#rental_start_time').val();
   raw_end_time = $('#rental_end_time').val();
@@ -45,6 +46,21 @@ function costCalculation(){
     });
    }
 }
+
+$('.selectable-item-type').click(function (e) {
+  if($(this).hasClass('active')) { //was active lets remove it
+    $(this).removeClass('active');
+    $(this).html($(this).attr('_name')); //set to base item type name
+  } else { //was inactive lets active it
+    //set item type to active
+    $(this).addClass('active');
+    //add input for ammount
+    $(this).html("<input type='number' step='1' value='1' class='item-type-quantity-picker cost-dependent'> " + $(this).attr('_name')); //set to base item type name
+  }
+
+  $('.cost-dependent').trigger('change'); //force a recalculation
+  e.preventDefault();
+});
 
 $(document).ready( function () {
   costCalculation(); //initial cost calculaton
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -23,12 +23,19 @@ class RentalsController < ApplicationController
   # GET /rentals/1
   def show; end
 
-  # GET /rentals/cost?end_time=time&start_time=time&item_type=1
+  # GET /rentals/cost?end_time=time&start_time=time&item_type_ids=...
   def cost
-    start_time = Time.zone.parse(params[:start_time]).to_date.to_s
-    end_time = Time.zone.parse(params[:end_time]).to_date.to_s
-    item_type = ItemType.find(params[:item_type]) if params[:item_type]
-    render json: Rental.cost(start_time, end_time, item_type)
+    _params = cost_params
+    required_params = %i(start_time end_time item_types)
+    if (_params & required_params) == required_params
+      start_time = Time.zone.parse(_params[:start_time]).to_date.to_s
+      end_time = Time.zone.parse(_params[:end_time]).to_date.to_s
+
+      item_type = ItemType.find(params[:item_type]) if params[:item_type]
+      render json: item_type.cost(start_time, end_time, item_type)
+    else
+      render json: {errors: { missing_params: (required_params - _params)} }
+    end
   end
 
   # GET /rentals/new
@@ -140,6 +147,10 @@ class RentalsController < ApplicationController
   def invoice; end
 
   private
+
+  def cost_params
+    params.permit(:start_time,:end_time, item_types: [])
+  end
 
   def set_users_to_assign
     @users = @current_user.assignable_renters.map do |user|
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -36,14 +36,14 @@ class RentalsController < ApplicationController
 
   # GET /rentals/cost?end_time=time&start_time=time&item_types=...
   def cost
-    _params = cost_params
+    cost_params = params.permit(:start_time, :end_time, item_types: [])
     required_params = %w(start_time end_time item_types)
-    if (_params.to_h.keys & required_params) == required_params
-      start_time = Time.zone.parse(_params[:start_time]).to_date.to_s
-      end_time = Time.zone.parse(_params[:end_time]).to_date.to_s
+    if (cost_params.to_h.keys & required_params) == required_params
+      start_time = Time.zone.parse(cost_params[:start_time]).to_date.to_s
+      end_time = Time.zone.parse(cost_params[:end_time]).to_date.to_s
 
       begin
-        cost = _params[:item_types].each_with_object({}) do |it_id, acc|
+        cost = cost_params[:item_types].each_with_object({}) do |it_id, acc|
           if it = ItemType.find_by(id: it_id)
             acc[it.name] ||= 0
             acc[it.name] += it.cost(start_time, end_time)
@@ -58,7 +58,7 @@ class RentalsController < ApplicationController
       render json: cost.merge(_total: cost.values.reduce(:+))
     else
       render status: 400, json: { errors: [
-        "missing_params: #{(required_params - _params.to_h.keys).inject('') { |acc, part| acc.blank? ? part.to_s : "#{acc}, #{part}" }}"
+        "missing_params: #{(required_params - cost_params.to_h.keys).inject('') { |acc, part| acc.blank? ? part.to_s : "#{acc}, #{part}" }}"
       ] }
     end
   end
@@ -174,10 +174,6 @@ class RentalsController < ApplicationController
   def invoice; end
 
   private
-
-  def cost_params
-    params.permit(:start_time, :end_time, item_types: [])
-  end
 
   def set_users_to_assign
     @users = @current_user.assignable_renters.map do |user|
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -46,8 +46,8 @@ class RentalsController < ApplicationController
 
       render json: cost.merge(_total: cost.values.reduce(:+))
     else
-      render json: { errors: [
-        "missing_params: #{(required_params - _params).inject('') { |acc, part| if acc.blank? then "#{part}" else "#{acc}, #{part}" end }}"
+      render status: 400, json: { errors: [
+        "missing_params: #{(required_params - _params.to_h.keys).inject('') { |acc, part| if acc.blank? then "#{part}" else "#{acc}, #{part}" end }}"
       ]}
     end
   end
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -52,7 +52,7 @@ class RentalsController < ApplicationController
           end
         end
       rescue => err
-        render(json: { errors: ["item not found #{err.message}"] }, status: 400) and return
+        render(json: { errors: ["item not found #{err.message}"] }, status: 400) && (return)
       end
 
       render json: cost.merge(_total: cost.values.reduce(:+))
@@ -138,7 +138,7 @@ class RentalsController < ApplicationController
 
     # if the specified a custom amount we will create the financial transactions later
     rental.skip_financial_transactions = params[:amount] && rental.cost != params[:amount] &&
-                                          @current_user.has_permission?('rentals', 'cost_adjustment')
+                                         @current_user.has_permission?('rentals', 'cost_adjustment')
 
     if rental.save
       if rental.skip_financial_transactions # use that special pricing, we already verified their perms
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -123,23 +123,25 @@ class RentalsController < ApplicationController
 
   # POST /rentals
   def create
-    @rental = Rental.new(rental_params.merge(creator: @current_user))
+    rentals_items = params[:rentals_items][:item_types].map do |it_id|
+      RentalsItem.new(item_type_id: it_id)
+    end
+    rental = Rental.new(rental_params.merge(creator: @current_user, rentals_items: rentals_items))
 
-    @start_date = params['start_date'] || Time.zone.today
-    if @rental.save
+    if rental.save
       if params[:amount] && @current_user.has_permission?('rentals', 'cost_adjustment')
+        # TODO this needs to be changed
         # find existing financial_transaction and change it
-        @financial_transaction = FinancialTransaction.find_by rental: @rental, transactable_type: Rental.name, transactable_id: @rental.id
-        @financial_transaction.amount = params[:amount]
-        @financial_transaction.save
+        financial_transaction = FinancialTransaction.find_by rental: rental, transactable_type: Rental.name, transactable_id: rental.id
+        financial_transaction.amount = params[:amount]
+        financial_transaction.save
       end # if they dont have permission ignore it and we will use default pricing
       flash[:success] = 'Rental Successfully Reserved'
-      redirect_to(@rental)
+      redirect_to(rental)
     else # error has problem, cannot rental a error message here
-      set_users_to_assign
-      flash[:warning] = (@rental.item_type.try(:name) || 'Item type') + ' Is Not Available For Specified Dates'
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
-      render :new
+      flash[:warning] = (rental.item_type.try(:name) || 'Item type') + ' Is Not Available For Specified Dates'
+      rental.errors.full_messages.each { |e| flash_message :warning, e }
+      redirect_to :new
     end
   end
 
@@ -202,9 +204,9 @@ class RentalsController < ApplicationController
   # Only allow a trusted parameter "white list" through.
   def rental_params
     # tokeninput gives us an array, but find_by doesnt care it just gives us the first one
-    user = User.find_by id: params.require(:rental).require(:renter_id)
+    renter = User.find_by id: params.require(:rental).require(:renter_id)
     new_time = Time.zone.parse(params[:rental][:end_time]).end_of_day
-    params.require(:rental).permit(:start_time, :item_type_id, :pickup_name, :dropoff_name,
-                                   :pickup_phone_number, :dropoff_phone_number).merge(renter: user, end_time: new_time)
+    params.require(:rental).permit(:start_time, :pickup_name, :dropoff_name,
+                                   :pickup_phone_number, :dropoff_phone_number).merge(renter: renter, end_time: new_time)
   end
 end
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -53,7 +53,7 @@ class RentalsController < ApplicationController
           end
         end
       rescue => err
-        render(json: { errors: ["item not found #{err.message}"] }, status: 400) && (return)
+        render(json: { errors: ["item not found #{err.message}"] }, status: 400) and return
       end
 
       render json: cost.merge(_total: cost.values.reduce(:+))
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -128,12 +128,14 @@ class RentalsController < ApplicationController
     rental = Rental.new rental_params.merge(creator: @current_user)
 
     if rental.save
+      # by deleting and createing a new ft we cause two emails to be sent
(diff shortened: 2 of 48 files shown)


app/assets/javascripts/rentals/new.js.erb
```diff
@@ -13,40 +13,47 @@ function days_between(sd, ed) {
...
+  var date_range = $('#date_range');
 
-  if(!isEmpty(raw_item_type) && !isEmpty(raw_end_time) && !isEmpty(raw_start_time)) {
+  if(!item_types.length <= 0 && !isEmpty(raw_end_time) && !isEmpty(raw_start_time)) {
     if(new Date(raw_start_time) > new Date(raw_end_time)) {
       return; //do not perform a cost calculation
       //the end_date changed to start_date + previous range then call this method once again
     }
-    range = days_between(raw_start_time, raw_end_time);
+    var range = days_between(raw_start_time, raw_end_time);
```
Anbranin: Are you using Moment.js? If not I highly recommend it. 
http://momentjs.com/docs/

app/controllers/rentals_controller.rb
```diff
@@ -23,20 +34,42 @@ def index
...
+        cost = _params[:item_types].each_with_object({}) do |it_id, acc|
+          if it = ItemType.find_by(id: it_id)
+            acc[it.name] ||= 0
+            acc[it.name] += it.cost(start_time, end_time)
+          else
+            raise ArgumentError, it_id
+          end
+        end
+      rescue => err
+        render(json: { errors: ["item not found #{err.message}"] }, status: 400) && (return)
```
Anbranin: When you do "and return" you actually want the word "and"

app/controllers/rentals_controller.rb
```diff
@@ -102,23 +135,26 @@ def update
...
       flash[:success] = 'Rental Successfully Reserved'
-      redirect_to(@rental)
-    else # error has problem, cannot rental a error message here
-      set_users_to_assign
-      flash[:warning] = (@rental.item_type.try(:name) || 'Item type') + ' Is Not Available For Specified Dates'
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
-      render :new
+      redirect_to(rental)
+    else
+      # flash[:warning] = 'Item type is not available for specified dates' # TODO find out the actual error
```
Anbranin: Why don't you know already what the actual error is?

app/controllers/rentals_controller.rb
```diff
@@ -23,20 +34,42 @@ def index
...
   end
 
   # GET /rentals/new
   def new
-    @rental = Rental.new
+    @rental ||= Rental.new # conditionally assign because it could be set already by a method that calls this one
     @start_date = params['start_date'].try(:to_date) || Time.zone.today
     @admin_status = @current_user.has_group? Group.where(name: 'admin')
     set_users_to_assign
+    render :new # this call is required because we directly call the new method on failure of create and other methods
```
Anbranin: It is actually not required to write this.
In the create method if you said `render :new` then the create method would render the `new` page, not call this controller action.

app/controllers/rentals_controller.rb
```diff
@@ -102,23 +135,26 @@ def update
...
-      set_users_to_assign
-      flash[:warning] = (@rental.item_type.try(:name) || 'Item type') + ' Is Not Available For Specified Dates'
-      @rental.errors.full_messages.each { |e| flash_message :warning, e, :now }
-      render :new
+      redirect_to(rental)
+    else
+      # flash[:warning] = 'Item type is not available for specified dates' # TODO find out the actual error
+      rental.errors.full_messages.each { |e| flash_message :warning, e }
+      @rental = rental
+      new && return # render new
```
Anbranin: Couple questions: First of all, does just writing the word "new" actually render the `new` page? Never seen that before.
Also, again, you don't want `&&` you want `and`

app/controllers/rentals_controller.rb
```diff
@@ -23,20 +34,42 @@ def index
...
   def show; end
 
-  # GET /rentals/cost?end_time=time&start_time=time&item_type=1
+  # GET /rentals/cost?end_time=time&start_time=time&item_types=...
   def cost
-    start_time = Time.zone.parse(params[:start_time]).to_date.to_s
-    end_time = Time.zone.parse(params[:end_time]).to_date.to_s
-    item_type = ItemType.find(params[:item_type]) if params[:item_type]
-    render json: Rental.cost(start_time, end_time, item_type)
+    _params = cost_params
```
dfaulken: You should only preface a variable with an underscore when you're not using it. This is used for when you're pattern-matching something into multiple variables, but only need one. For instance:

```ruby
a, _b = some_method_that_returns_a_tuple

hash.each_pair do |key, _value|
  i_only_want_the(key)
end
```

app/controllers/rentals_controller.rb
dfaulken: Yes, and also, there's no reason to write parentheses around a single word.

app/controllers/rentals_controller.rb
dfaulken: Right, you have validations in the model.

app/controllers/rentals_controller.rb
dfaulken: My guess would be that it instantiates a new instance of the controller.

app/helpers/application_helper.rb
```diff
@@ -79,10 +79,10 @@ def time_field(object_name, method, options = {})
   def flash_message(type, text, now = nil)
     if now
       flash.now[type] ||= []
-      flash.now[type] << text
+      flash.now[type] << text.dup
```
dfaulken: Why?

app/models/hold.rb
```diff
@@ -19,22 +19,21 @@ class Hold < ActiveRecord::Base
   def handle_conflicting_rentals
     # if rental falls between data range of our hold
     # only reserved but not picked up
-    conflicting_rentals = Rental.reserved.where('start_time <= :hold_end_time AND end_time >= :hold_start_time AND item_id = :hold_item_id',
-                                                hold_start_time: start_time, hold_end_time: end_time, hold_item_id: item.id)
+    conflicting_rentals = Rental.reserved.joins('INNER JOIN rentals_items ON rentals_items.rental_id = rentals.id')
+                                .where('start_time <= :hold_end_time AND end_time >= :hold_start_time AND rentals_items.item_id = :hold_item_id',
```
dfaulken: Interesting, I didn't know you could keyword-interpolate.

I'd suggest a model method for this.

app/models/hold.rb
```diff
@@ -19,22 +19,21 @@ class Hold < ActiveRecord::Base
...
     start_hold
   end
 
   def replace_rental(curr_rental)
-    new_rental = Rental.new(item_type_id: curr_rental.item_type_id, creator_id: curr_rental.creator_id,
-                            renter_id: curr_rental.renter_id,
-                            start_time: curr_rental.start_time, end_time: curr_rental.end_time)
-    new_rental.create_reservation
+    new_rental = Rental.new(rentals_items_attributes: (curr_rental.item_types.map { |x| { item_type: x } }), creator_id: curr_rental.creator_id,
+                            renter_id: curr_rental.renter_id, start_time: curr_rental.start_time, end_time: curr_rental.end_time)
```
dfaulken: If `creator` and `renter` are named associations, you should be able to reference them without the ID: `creator: curr_rental.creator, renter: curr_rental.renter`.

app/models/hold.rb
```diff
@@ -54,7 +53,8 @@ def lift_hold
 
   def conflicting_ongoing_rental
     # it couldnt be possible that the same item has already been picked up twice under the same rental
-    Rental.picked_up.find_by('start_time <= :hold_end_time AND end_time >= :hold_start_time AND item_id = :hold_item_id',
-                             hold_start_time: start_time, hold_end_time: end_time, hold_item_id: item.id)
+    Rental.picked_up.joins('INNER JOIN rentals_items ON rentals_items.rental_id = rentals.id')
```
dfaulken: Again, I'd suggest a model method.

app/models/hold.rb
```diff
@@ -54,7 +53,8 @@ def lift_hold
...
+          .find_by('start_time <= :hold_end_time AND end_time >= :hold_start_time AND rentals_items.item_id = :hold_item_id',
+                   hold_start_time: start_time, hold_end_time: end_time, hold_item_id: item.id)
```
dfaulken: Where are these local variables coming from?

app/models/item_type.rb
```diff
@@ -4,4 +4,32 @@ class ItemType < ActiveRecord::Base
...
+    # Long term pricing
+    longterm_prices = longterm_cost(no_of_weeks)
+
+    unless longterm_prices.nil? # catch cases where item is neither 4 nor 6 seat cart
+      return longterm_prices[no_of_weeks] + ((rental_duration % 7) * fee_per_day)
+    end
+    (days_to_charge_for * fee_per_day) + base_fee
+  end
+
+  def longterm_cost(weeks)
```
dfaulken: This seems like a *terrible* place to encode cost information. If there are more kinds of carts, you have to change your code. If you ever want to make changes to your costs (which you will), you have to change your code.

Ideally the people who care about money would be able to change the costs themselves, I imagine.

app/models/rental.rb
```diff
@@ -8,18 +8,24 @@ class Rental < ActiveRecord::Base
   has_many :financial_transactions
   has_one :financial_transaction, as: :transactable
 
-  after_create :create_financial_transaction
+  # will skip financial transactions if we plan to create manual pricing
+  attr_accessor :skip_financial_transactions
+  after_create :create_financial_transaction, unless: proc { |rental| rental.skip_financial_transactions }
```
dfaulken: Does `unless: :skip_financial_transactions` not work?

app/models/rental.rb
```diff
@@ -8,18 +8,24 @@ class Rental < ActiveRecord::Base
   has_many :financial_transactions
   has_one :financial_transaction, as: :transactable
 
-  after_create :create_financial_transaction
+  # will skip financial transactions if we plan to create manual pricing
+  attr_accessor :skip_financial_transactions
```
dfaulken: Is there the possibility of a race condition with this?

app/models/rental.rb
```diff
@@ -40,12 +46,24 @@ def renter_is_assignable
...
   scope :with_balance_due, -> { Rental.where id: Rental.select { |rental| rental.balance.positive? }.collect(&:id) }
-  scope :with_balance_over, ->(min) { Rental.where id: Rental.select { |rental| rental.balance >= min }.collect(&:id) }
+  scope :with_balance_over, ->(min) { Rental.where id: Rental.select { |rental| rental.balance > min }.collect(&:id) }
 
   delegate :payments, to: :financial_transactions
   delegate :department, to: :renter
 
+  def reservations
+    rentals_items.collect(&:reservation_id)
+  end; alias reservation_ids reservations
```
dfaulken: Why write this on the same line as the `end`?

app/models/item_type.rb
```diff
@@ -4,4 +4,32 @@ class ItemType < ActiveRecord::Base
...
+    # prices are specially defined
+    if weeks >= 2
+      if name == '4 Seat'
+        { 2 => 500, 3 => 700, 4 => 850 }
+      elsif name == '6 Seat'
+        { 2 => 600, 3 => 900, 4 => 1100 }
+      end
+    end
+  end; private :longterm_cost
```
dfaulken: Why write this on the same line as the `end`?

app/models/rental.rb
```diff
@@ -40,12 +46,24 @@ def renter_is_assignable
...
+  #   if these queries ever become a problem this would be faster, i think a cross apply or something like that would probably be even faster.
+  #   select * from rentals where
+  #       ((select sum(amount) from financial_transactions where rental_id=rentals.id and
+  #         (transactable_type='Payment' or transactable_type='Cancelation'))
+  #       -(select sum(amount) from financial_transactions where rental_id=rentals.id and
+  #         not (transactable_type='Payment' or transactable_type='Cancelation')))
+  #       > 0
   scope :with_balance_due, -> { Rental.where id: Rental.select { |rental| rental.balance.positive? }.collect(&:id) }
-  scope :with_balance_over, ->(min) { Rental.where id: Rental.select { |rental| rental.balance >= min }.collect(&:id) }
+  scope :with_balance_over, ->(min) { Rental.where id: Rental.select { |rental| rental.balance > min }.collect(&:id) }
```
dfaulken: So, can I ask, as a design decision, why this PR's code passes around objects' IDs after every query instead of just the objects themselves?

app/models/rental.rb
```diff
@@ -92,34 +110,77 @@ def renter_is_assignable
...
+        created_reservations << reservation[:uuid]
+        ri.item = Item.find_by(name: reservation[:item][:name])
+      end
     rescue => error
-      errors.add :base, error.inspect
+      # roll back the rentals we got partially through creating
+      failed_roll_back = false
+      created_reservations.each do |uuid|
+        # remove reservation from RentalsItem
+        if (ri = rentals_items.to_a.find { |_ri| _ri.reservation_id == uuid })
```
dfaulken: It's better not to do variable assignment in a condition, but to do it before and then just write your condition using that variable.

app/models/rental.rb
```diff
@@ -92,34 +110,77 @@ def renter_is_assignable
...
-      self.reservation_id = nil
-    rescue => error
-      errors.add(:base, error.inspect) && (return false)
+    rentals_items.each do |ri|
+      next if ri.reservation_id.nil? # nothing to delete here
+      errors.add(:rentals_items, "Failed to delete reservation (uuid #{ri.reservation_id})") unless delete_reservation(ri.reservation_id)
+      ri.reservation_id = nil
     end
-    true
+    throw(:abort) if errors.any?
```
dfaulken: From what does this abort?

app/models/rental.rb
```diff
@@ -92,34 +110,77 @@ def renter_is_assignable
...
+  end
+
+  def delete_reservation(uuid)
+    Inventory.delete_reservation(uuid) # will throw errors if it fails
+  rescue
+    return false
+  end
+
+  def str_reservation_ids
+    rentals_items.reduce('') { |acc, part| "#{acc}, #{part.reservation_id}" }[2..-1]
```
dfaulken: This (and the two methods below it) definitely need comments that give an example of what they do.

app/models/rental.rb
```diff
@@ -176,35 +235,15 @@ def duration
...
+      rental_amount = ri.item_type.cost(start_time.to_date, end_time.to_date)
+      FinancialTransaction.create rental: self, amount: rental_amount, transactable_type: RentalsItem.name, transactable_id: ri.id
     end
   end
 
-  def create_financial_transaction
-    rental_amount = Rental.cost(start_time.to_date, end_time.to_date, item_type)
-    FinancialTransaction.create rental: self, amount: rental_amount, transactable_type: self.class, transactable_id: id
+  def cost
+    rentals_items.reduce(0) { |acc, part| acc + part.item_type.cost(start_time.to_date, end_time.to_date) }
```
dfaulken: I think `sum` would be easier than `reduce(0)`.

app/views/invoice_mailer/invoice.html.erb
```diff
@@ -47,86 +41,107 @@
...
-          <div class="text-right">
-            <h2><label>Total Amount Due:</label>
-            <%= number_to_currency @rental.balance %></h2>
+          </thead>
+          <tbody>
+            <% @rental.financial_transactions.each do |ft| %>
+              <tr>
+                <td><%= ft.created_at %></td>
+                <td><%= ft.transactable_type %></td>
+                <td><%= number_to_currency (ft.amount || 0) %></td>
```
dfaulken: No space between a method name and the parentheses that wrap its arguments.

app/views/payment_tracking/index.html.erb
```diff
@@ -32,7 +32,7 @@
       <tr>
         <th class="min-width nosortcol">Select All <input type="checkbox" id="payment_tracking_select_all"></th>
         <th>Rental Id</th>
-        <th>Reservation Id</th>
+        <th>Reservation Ids</th>
```
dfaulken: Not 'IDs'?

app/views/payment_tracking/index.html.erb
```diff
@@ -45,7 +45,7 @@
         <tr>
           <td><input type="checkbox" id=<%= "paymentTrackingCheck#{rental.id}" %> ></td>
           <td><%= rental.id %></td>
-          <td><%= rental.reservation_id %></td>
+          <td><%= rental.str_reservation_ids.truncate(77) %></td>
```
dfaulken: Why 77?

app/views/rentals/_form.html.erb
```diff
@@ -1,85 +1,87 @@
...
-  <div class="form-group col-xs-12">
-    <%= label_tag :date_range, 'Number of Days', class: 'control-label col-xs-2' %>
     <div class="col-xs-6">
-      <%  #(:start_time - :end_time) %>
-      <%= number_field_tag :date_range, 1, min: 0, disabled: true, class: 'form-control' %>
+      <div class="list-group">
+        <% @item_types.each do |it| %>
+          <div class="input-group selectable-item-type">
+            <%= link_to it.name, '#', class: 'list-group-item' %>
+            <!-- we use _name attr because any ele with a name attr is submitted to server, this is a dummy field, we have to fix it latter so it submitts properly -->
```
dfaulken: You're aware that HTML comments will be rendered in your DOM and inspectable by your users?

app/views/rentals/invoice.html.erb
```diff
@@ -1,120 +1,104 @@
...
+  <div class="col-xs-6">
+    <p>
+    Parking Services <br>
+    University of Massachusetts Amherst<br>
+    51 Forestry Way Amherst, MA 01003-9262<br>
+    phone: (413) 577-PARK <br>
+    fax: (413) 545-4440<br>
+    web: http://parking.umass.edu<br>
+    email: parking@admin.umass.edu<br>
+    </p>
```
dfaulken: I see this is repeated. Might be worth extracting into a partial incase any of the information changes. For instance, down the line you may wish to have your URL be `https` 😄

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.6
+    "covered_percent": 98.48
```
dfaulken: Hmm, so are you introducing untested code?

spec/controllers/rentals_controller_spec.rb
```diff
@@ -47,11 +48,46 @@
   end
 
   describe 'GET #cost' do
-    it 'returns a cost based on item type' do
-      start_time = Date.today
-      end_time = Date.tomorrow
-      get :cost, params: { item_type: item_type, start_time: start_time, end_time: end_time }
-      expect(response.body).to eq(Rental.cost(start_time, end_time, item_type).to_s)
+    context 'retrieves cost' do
```
dfaulken: A lot of repetition in these tests. Up to you as a team to decide whether you prefer that or prefer some abstraction with `let` statements and the like.

spec/models/hold_spec.rb
```diff
@@ -64,10 +64,17 @@
...
-    let(:hold) { create(:hold, item: shared_item) }
-    let(:conflicting_rental) { create(:hold_conflicting_rental, item: shared_item) }
-    let(:future_rental) { create(:far_future_rental, item: shared_item) }
+    let(:hold) { create(:hold, item: shared_item, start_time: 1.day.from_now, end_time: 10.days.from_now) }
+    let(:conflicting_rental) do
+      create(:rental, start_time: 2.days.from_now, end_time: 4.days.from_now,
+                      rentals_items: [build(:rentals_item, item: shared_item)])
+    end
+    let(:future_rental) do
+      create(:rental, start_time: 40.days.from_now, end_time: 42.days.from_now,
```
dfaulken: Hmm, I didn't know about `from_now`. We've always used `since`, which I suspect is an alias.

spec/models/item_type_spec.rb
```diff
@@ -20,4 +20,10 @@
   it 'is invalid with a negative fee_per_day' do
     expect(build(:item_type, fee_per_day: -1)).not_to be_valid
   end
+
+  describe '#cost' do
```
dfaulken: Is it worth doing any further testing than just "it's a number"?

spec/models/rental_spec.rb
```diff
@@ -159,6 +167,43 @@
...
+      it 'adds to errors if it fails to delete reservation' do
+        # will try to roll back but we wont let it
+        allow(Inventory).to receive(:delete_reservation).and_raise(StandardError)
+
+        r = build(:rental, rentals_items: [build(:rentals_item, reservation_id: nil), fail_rentals_item])
+        expect do
+          r.save!
+        end.to(raise_error(ActiveRecord::RecordNotSaved)) && change(Rental, :count).by(0) && change(RentalsItem, :count).by(0)
+        messages = r.errors.messages[:base]
+        expect(messages.count).to eq 2
```
dfaulken: BTW, you can write `to be 2`. There's only one `2`.

dfaulken: (`Fixnum`s are 'non-duplicable')

spec/models/rental_spec.rb
```diff
@@ -173,12 +218,13 @@
...
-    end
-    it 'returns false if the item is in_valid except for a missing reservation_id' do
-      expect(build(:invalid_rental, reservation_id: nil).valid?).to be false
+  describe '#cost' do
+    it 'gets the right cost based on item types' do
+      it1 = create :item_type, base_fee: 5, fee_per_day: 2
+      it2 = create :item_type, base_fee: 3, fee_per_day: 11
+      rental = build :rental, start_time: Time.zone.today, end_time: Time.zone.tomorrow,
+                              rentals_items_attributes: [{ item_type: it1 }, { item_type: it2 }]
+      expect(rental.cost).to eq((5 + 2 * 2) + (3 + 2 * 11))
```
dfaulken: Might be worth explaining the logic encoded in this numerical expression.

spec/models/rental_spec.rb
```diff
@@ -357,80 +424,74 @@
...
-    rent = create :mock_rental, item_type: six_seat, end_time: (Time.current + 20.days)
-    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([900])
-  end
-  it 'creates a 6 Seat 28 day financial transaction(longterm 4 week)' do
-    rent = create :mock_rental, item_type: six_seat, end_time: (Time.current + 27.days)
-    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([1100])
+  context 'rentals and financial transactions' do
+    it 'creates a 1 day financial transaction' do
+      rent = create :mock_rental, end_time: (Time.current + 1.second)
+      expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([110])
```
dfaulken: Is this equivalent to

```ruby
expect(rent.financial_transactions.first.amount).to eql 110`
```
?

spec/models/rental_spec.rb
```diff
@@ -357,80 +424,74 @@
...
-    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([600])
-  end
-  it 'creates a 6 Seat 21 day financial transaction(longterm 3 week)' do
-    rent = create :mock_rental, item_type: six_seat, end_time: (Time.current + 20.days)
-    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([900])
-  end
-  it 'creates a 6 Seat 28 day financial transaction(longterm 4 week)' do
-    rent = create :mock_rental, item_type: six_seat, end_time: (Time.current + 27.days)
-    expect(FinancialTransaction.where(rental: rent).map(&:amount)).to eq([1100])
+  context 'rentals and financial transactions' do
```
dfaulken: I think there's too much going on in these tests to express them all as just two lines of code, particularly the later tests.

app/assets/javascripts/rentals/new.js.erb
tomecho: I know we are using moment.js just not here, I could rewrite this method but id rather have it in a different pull request.  I was just adding the `var` keyword to keep the `range` variable within scope.

app/controllers/rentals_controller.rb
tomecho: Yeah it's a little weird but it does actually render the new page.
`spec/controllers/rentals_controller_spec.rb:192`

Also just wondering does `and` versus `&&` actually make a difference in this case or is it just a styling standard?

app/controllers/rentals_controller.rb
tomecho: in this case isn't `&&` is equivalent to `and`?

dfaulken: I mean yes, in this case, because you wrapped everything in parentheses. But they're for different things. The logical `&&` is for, well, logic — like in a conditional. `and` is for control flow.

app/controllers/rentals_controller.rb
tomecho: I added the line below which prints out all the errors,  I was questioning why this flash warning had such a specific error message in the past.  I could probably just delete it but I have the suspicion it does not recognize failed reservations.

app/controllers/rentals_controller.rb
dfaulken: I'm with @Anbranin on this one. No reason not to write an explicit render call. Why have something that's weird *but* works, when you could have something that works and isn't weird?

app/controllers/rentals_controller.rb
dfaulken: > I added the line below which prints out all the errors

So then this isn't a TODO? It's done?

app/assets/javascripts/rentals/new.js.erb
```diff
@@ -13,40 +13,47 @@ function days_between(sd, ed) {
   return Math.round(diff_ms/one_day) + 1;
 }
 
+function item_types_to_arr() {
+  //collect item types (will always be at least [])
+  return $(".item-type-quantity-picker").toArray().reduce(function (acc,part) {
```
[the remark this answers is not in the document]
tomecho: sorry the explanation is on the method that calls this one, `app/assets/javascripts/new.js.erb:84`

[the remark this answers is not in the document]
dfaulken: I read that. I'm still not clear on why things are being pushed a *dynamic* number of times.

app/helpers/application_helper.rb
tomecho: I was trying to debug a frozen string literal error in the end I dont even think it was this, good catch, ill remove this.

app/models/hold.rb
tomecho: instance variables

dfaulken: Oh, they're attributes of the model. I got so confused, I forgot I was looking at an instance method in a model file.

app/models/item_type.rb
tomecho: It was just a copy paste from `app/models/rentals.rb`

app/models/rental.rb
tomecho: Could you provide an example?  I am not certain I understand.

app/models/rental.rb
tomecho: In this case I think its because I wanted an active record relation not an array.  This could easily be rewritten it is already inefficient as it is.

app/models/rental.rb
tomecho: https://github.com/umts/aggressive-epsilon/commit/cda7c968cac6d67d679a1e8fb490b8d0c52fa18a

I figured it was the right way to do things

app/models/rental.rb
tomecho: `rental.destroy`

app/views/payment_tracking/index.html.erb
tomecho: that is two uuid's separated by a space and a comma with 3 extra characters for the ... to show truncation.

app/views/rentals/_form.html.erb
tomecho: yes, they can also find this project open source on github

app/models/rental.rb
dfaulken: Hmm, Rubocop must've suggested it. Carry on.

app/views/rentals/invoice.html.erb
tomecho: Do you mean repeated across the two invoice pages, that will be solved in #359

coverage/.last_run.json
tomecho: shouldn't be, i am not certain how this is calculated but perhaps the total number of lines has increased

app/models/rental.rb
dfaulken: Your code changes the class-wide attr_accessor, and another application instance accesses the DB at the same time.

coverage/.last_run.json
tomecho: looks like there might be a couple lines that were not tested

app/views/rentals/_form.html.erb
dfaulken: True. It's also just additional bulk in your HTML that's really just meant internally for your developers. It's not a huge security concern, but doesn't seem like what you want. Particularly there are correctly done ERB comments on this very page.

app/views/payment_tracking/index.html.erb
dfaulken: Okay. Since all of that logic is encoded in your method, I'd suggest having that truncation be an option you can specify in the method arguments.

spec/models/rental_spec.rb
tomecho: Should be yes

app/models/rental.rb
dfaulken: Okay. It'd be nice to explain that here.

app/models/item_type.rb
dfaulken: I mean, okay. Still seems like a bad idea.

app/controllers/rentals_controller.rb
```diff
@@ -23,20 +34,42 @@ def index
...
+    required_params = %w(start_time end_time item_types)
+    if (_params.to_h.keys & required_params) == required_params
```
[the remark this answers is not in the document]
tomecho: this checks the intersection meaning they could pass more than the required params but seeing as the `cost_params` method only permits those three params lines 40 and 41 are now redundant

app/controllers/rentals_controller.rb
tomecho: I left that line there because of my suspicion it does not recognize failed reservations because of items not available.

app/assets/javascripts/rentals/new.js.erb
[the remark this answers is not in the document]
tomecho: because of the way im constructing the parameters to `Rental.new` I need rentals_items_attributes in a way such that multiple item types yield multiple rentals_items not just a single one.  I could have this logic here or in the controller, I just chose to have it client side.

app/models/item_type.rb
tomecho: It is, I agree with you and I have created a separate issue to deal with this #361

app/models/rental.rb
tomecho: oh shoot, I thought attr_accessors values were instance variables

dfaulken: They are. I'm just curious if they're model wide like that, so that while your code has it set one way or the other, it could affect other people interacting with the model. I'm honestly not sure.

app/controllers/rentals_controller.rb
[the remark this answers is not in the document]
tomecho: oh actually i remember why i did that, params.permit doesnt require the presence of variables, and that is what i wanted to do.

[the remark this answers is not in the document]
dfaulken: > params.permit doesnt require the presence of variables

`params.require` does

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.6
+    "covered_percent": 68.54
```
anmerkl: That's a huge drop.  Did you not run the entire test suite?

app/controllers/rentals_controller.rb
```diff
@@ -36,14 +36,14 @@ def show; end
 
   # GET /rentals/cost?end_time=time&start_time=time&item_types=...
   def cost
-    _params = cost_params
+    cost_params = params.permit(:start_time, :end_time, item_types: [])
```
anmerkl: If you change this to `params.require` like Dave mentioned, would this get rid of the need of `cost_params.to_h.keys & required_params` ?
My main concern is code readability.  If your way is the best way, then I'm totally fine with it.

coverage/.last_run.json
tomecho: no

app/controllers/rentals_controller.rb
tomecho: The only thing my method actually has over params.require is that it allows me to form a response in my own format if the param is missing.

app/models/rental.rb
tomecho: Could you possibly give an example, I am not sure I understand what you're saying.

dfaulken: The more I think about it, the more I suspect it's not a concern. But presumably you're running this server with something like Phusion Passenger, which makes available multiple running instances of your app so that you can serve multiple requests simultaneously. So, my concern was that while one instance of the app was processing a request which involved setting the model-wide `attr_accessor` to skip financial transactions, another app instance responding to another, unrelated request involving the same model might also skip financial transactions.

But, I think since it's a separate instance of the app running, and that `attr_accessor` is essentially a value in the running memory of a single instance, that it's not a real concern, since a single app instance can't serve multiple requests at once.

@werebus, is that correct?

werebus: You are correct, I don't think it's a concern.

Although they are the same _database_ object, it's very unlikely (impossible? probably) that you have two references to the same _ruby_ object.

PR #370, opened 2017-12-04
Fix processing page

### LINKED ISSUE #368: rentals#processing is broken
It wasnt ready for rentals has many rentals_items

tomecho:
fixes #368 
and a bug where the search form doesnt work at all, it just redirects to regular rentals index

### DIFF
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -92,7 +92,18 @@ class RentalsController < ApplicationController
 
   # GET /rentals/processing
   def processing
-    @q = rentals_visible_to_current_user.search(params[:q])
+    # first pull out the rentals that match our two fields that are not done through ransack
+    # (because arel is honestly just too complicated)
+    base_search_area = rentals_visible_to_current_user.joins(:rentals_items)
+    if params[:item_type_id_in].present?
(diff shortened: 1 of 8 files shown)


app/controllers/rentals_controller.rb
```diff
@@ -92,7 +92,18 @@ def training_pdf
 
   # GET /rentals/processing
   def processing
-    @q = rentals_visible_to_current_user.search(params[:q])
+    # first pull out the rentals that match our two fields that are not done through ransack
+    # (because arel is honestly just too complicated)
+    base_search_area = rentals_visible_to_current_user.joins(:rentals_items)
+    if params[:item_type_id_in].present?
+      base_search_area = base_search_area.where('rentals_items.item_type_id' => params.permit(item_type_id_in: []).fetch(:item_type_id_in))
```
tomecho: @dfaulken do you know of a way to make this a fancy ransack matcher instead?  I'd like to have a custom ransacker that joins rentals to rentals_items and then checks if any rentals_items for the rental has a item_type_in some set.

dfaulken: I do not. I haven't used ransack before.

PR #373, opened 2017-12-08
Cancel rentals fix

### LINKED ISSUE #372: cannot cancel rentals
looks like canceling rentals runs into an error.  although it seems like it deletes in reservation api.

tomecho:
fixes #372

### DIFF
--- a/app/models/rental.rb
+++ b/app/models/rental.rb
@@ -16,7 +16,7 @@ class Rental < ActiveRecord::Base
   before_save :create_reservations, unless: proc { |rental| rental.reservation_ids.any? }
 
   # unreserve the items
-  before_destroy :delete_reservations
+  before_destroy :before_destroy_delete_reservations
 
   has_many :rentals_items, dependent: :destroy, inverse_of: :rental
   accepts_nested_attributes_for :rentals_items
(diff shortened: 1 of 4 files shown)


app/models/rental.rb
```diff
@@ -168,6 +167,12 @@ def delete_reservation(uuid)
     return false
   end
 
+  # same as delete_reservations but it aborts the destroy
+  def before_destroy_delete_reservations
+    delete_reservations # will add errors to obj
+    throw(:abort) if errors.any? # abort a #destroy
```
Anbranin: Would it be possible to use a [transaction](http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) instead?

tomecho: I don't think they are interchangeable.  The throw abort is to abort a ActiveRecord destroy action.  However now that I think of it, I think rails won't even complete the destroy unless errors are empty.

tomecho:
this is @Cobchise's problem now

app/models/rental.rb
tomecho: jsut checked and it is necessary to explicitly say `throw(:abort)`, just having the errors won't do anything.

Anbranin:
@Cobchise do you think this is safe to merge, or did you want to review it and then merge? Either way, can we close this at some point?

Cobchise:
I don't see the need to merge this, I believe the problem was solved with #379. I'll chat through why on Monday

PR #383, opened 2018-02-12
Fixed references to old model relations 'rental.item_type' and 'rental.item'

### LINKED ISSUE #377: Rental Item Types
Rentals currently have a has_many relation to both items and item_types. I assume this behavior was added to allow the creation of a rental that involves multiple different items/item types as opposed to the initial behavior that a rental was just a single item, however it seems to be only half implemented.

For example /incurred_incidentals is throwing an error as there is old code that assumes rentals only have one item type. I intend on fixing these errors by looping through each of the rentals' possible item types, and assume the above is correct.

Cobchise:
### Closes #377 

Ran a search through the codebase for `rental.item_type` and `rental.item` and replaced them with `rental.str_item_types` and `rental.str_items` accordingly (two pre-existing model methods).

### DIFF
--- a/app/views/incurred_incidentals/index.html.erb
+++ b/app/views/incurred_incidentals/index.html.erb
@@ -10,7 +10,7 @@
   </tr>
   <% @incurred_incidentals.each do |incurred_incidental| %>
     <tr>
-      <td><%= incurred_incidental.rental.item_type.name %></td>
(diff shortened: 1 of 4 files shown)


anmerkl:
A quick grep shows many more `rental.item_type` and `rental.item` than the 4 that you changed.  What's the reasoning behind this change in general and changing just the instances that you changed

PR #385, opened 2018-02-12
Integrate persistent active job backend

### LINKED ISSUE #374: Rental#create is taking a long time
It seems like the problem is because of the views taking forever to render, i'd bet at least part of that is the mail, we are already delivering async with deliver_later but it seems like it still tries to render right away. I think the reason is explained here, https://stackoverflow.com/a/28780106/4463703

`Completed 200 OK in 30894ms (Views: 30733.9ms | ActiveRecord: 115.1ms)`

Sucker punch seems like it would work fine for our application and is self contained which makes things simple.  https://github.com/brandonhilkert/sucker_punch

Installing is easy too just add this to the config
`config.active_job.queue_adapter = :sucker_punch`

Cobchise:
### Closes #374 

Installed sidekiq gem. Works fine in development, will need extra work in production. Both sidekiq and redis need to be running alongside the app. There are a number of ways to achieve this including [the capistrano gem](https://github.com/seuros/capistrano-sidekiq) or making our own scripts on the production server. 

[Useful link](https://github.com/mperham/sidekiq/wiki/Deployment)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ before_script:
   - bundle exec rake db:migrate
   - bundle exec rake db:seed
   - export TZ="America/New_York"
+  - bundle exec sidekiq -d -L /tmp/sidekiq.log
+services:
+  - redis-server
 addons:
   code_climate:
     repo_token: 15ed3ea3aa5e152f478928dbb01e5747253e54be2b90553170931dbc842428c7
--- a/Capfile
+++ b/Capfile
@@ -24,6 +24,7 @@ require 'capistrano/rails'
 require 'capistrano/rails/assets'
 require 'capistrano/rails/migrations'
 require 'capistrano/passenger'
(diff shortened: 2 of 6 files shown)


anmerkl:
Is this good to go?
Did you finish the "extra work in production"

Cobchise:
The code is good is to go, I'll set up the production side stuff and then ping here when that's complete

Anbranin:
What kind of stuff needs to be done on the server? Please make sure you document what you did / are doing so more than one person knows how the server is configured. :)
Also, you may want to add the [capistrano sidekiq gem](https://github.com/seuros/capistrano-sidekiq) if you want capistrano to restart sidekiq.

Cobchise:
https://blog.sensible.io/2013/08/20/setting-up-redis-for-production-environment.html

PR #397, opened 2018-02-28
Form and Model Validation Cleanup

### LINKED ISSUE #394: Form Validations
Currently, only some forms have specific validations on its fields (correctly formatted phone number, email, numbers > 0, etc.).  There are several forms that don't have crucial validations in place. For example, the Pick Up/Drop off forms accept any string as a phone number.

Form validations should consistently be applied wherever applicable.

anmerkl:
closes #394 

- Added validations to models
- Made sure to enforce validations in views where necessary
- Cleaned up views a little

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -30,14 +30,19 @@ engines:
     enabled: true
   fixme:
     enabled: true
+    checks:
+      XXX:
+        enabled: false
   rubocop:
     enabled: true
     checks:
       Rubocop/Style/SingleLineBlockParams:
         enabled: false
-      Rubocop/Rails/DynamicFindBy:
-        enabled: false
       Rubocop/Style/AndOr:
+        enabled: false
+      Rubocop/Style/IfUnlessModifier:
+        enabled: false
+      Rubocop/Rails/DynamicFindBy:
         enabled: false
       Rubocop/Metrics/BlockLength:
         enabled: false
--- a/app/assets/javascripts/financial_transactions/new.js
+++ b/app/assets/javascripts/financial_transactions/new.js
@@ -1,9 +1,9 @@
 $(document).ready(function() {
   $('#contact_email').change(function (){
-    var email_regex = /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/i;
+    var email_regex = /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i;
     if(!email_regex.test($('#contact_email').val()))
       $('#contact_email').parent().parent().addClass('has-error');
-    else 
+    else
       $('#contact_email').parent().parent().removeClass('has-error');
   });
 
@@ -11,7 +11,7 @@ $(document).ready(function() {
     var email_regex = /^[0-9]{10}$/i;
     if(!email_regex.test($('#contact_phone').val()))
       $('#contact_phone').parent().parent().addClass('has-error');
-    else 
+    else
       $('#contact_phone').parent().parent().removeClass('has-error');
   });
 });
--- a/app/assets/javascripts/financial_transactions/new.js
+++ b/app/assets/javascripts/financial_transactions/new.js
@@ -1,17 +1,9 @@
 $(document).ready(function() {
   $('#contact_email').change(function (){
-    var email_regex = /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i;
+    var email_regex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/i;
     if(!email_regex.test($('#contact_email').val()))
       $('#contact_email').parent().parent().addClass('has-error');
     else
       $('#contact_email').parent().parent().removeClass('has-error');
-  });
-
-  $('#contact_phone').change(function (){
-    var email_regex = /^[0-9]{10}$/i;
-    if(!email_regex.test($('#contact_phone').val()))
-      $('#contact_phone').parent().parent().addClass('has-error');
-    else
-      $('#contact_phone').parent().parent().removeClass('has-error');
   });
 });
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -2,9 +2,14 @@
 class Payment < ApplicationRecord
   enum payment_type: [:recharge]
   has_one :financial_transaction, as: :transactable
+
+  before_validation do
+    self.contact_phone = contact_phone.gsub(/\W/, '') if attribute_present? "contact_phone"
+  end
+
   validates :payment_type, :contact_name, :contact_phone, :contact_email, presence: true
-  validates :contact_phone, format: { with: /\d{10}/,
-                                      message: 'Phone number should be 10 digits' }
-  validates :contact_email, format: { with: /[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}/,
+  validates :contact_phone, format: { with: /\d{8,}/,
+                                      message: 'Phone number should not contain any letters and be at least 8 digits long' }
+  validates :contact_email, format: { with: /[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}/,
                                       message: 'Email improperly formatted'}
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -4,12 +4,12 @@ class Payment < ApplicationRecord
   has_one :financial_transaction, as: :transactable
 
   before_validation do
-    self.contact_phone = contact_phone.gsub(/\W/, '') if attribute_present? "contact_phone"
+    self.contact_phone = contact_phone.gsub(/\W/, '') if attribute_present? 'contact_phone'
   end
 
   validates :payment_type, :contact_name, :contact_phone, :contact_email, presence: true
-  validates :contact_phone, format: { with: /\d{8,}/,
+  validates :contact_phone, format: { with: /\A\d{8,}\z/,
                                       message: 'Phone number should not contain any letters and be at least 8 digits long' }
-  validates :contact_email, format: { with: /[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}/,
-                                      message: 'Email improperly formatted'}
+  validates :contact_email, format: { with: /\A[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}\z/,
+                                      message: 'Email improperly formatted' }
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 class Payment < ApplicationRecord
-  has_one :financial_transaction, as: :transactable
   enum payment_type: [:recharge]
-  validates :payment_type, :contact_name, :contact_email, :contact_phone, presence: true
-  validates :contact_phone, length: { is: 10 }
+  has_one :financial_transaction, as: :transactable
+  validates :payment_type, :contact_name, :contact_phone, :contact_email, presence: true
+  validates :contact_phone, format: { with: /\d{10}/,
+                                      message: 'Phone number should be 10 digits' }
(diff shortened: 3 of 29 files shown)


app/assets/javascripts/financial_transactions/new.js
```diff
@@ -1,17 +1,17 @@
 $(document).ready(function() {
   $('#contact_email').change(function (){
-    var email_regex = /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/i;
+    var email_regex = /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i;
```
Anbranin: I'd definitely suggest JQuery's [validator](https://jqueryvalidation.org/) at some point.

app/views/incidental_types/_form.html.erb
```diff
@@ -1,45 +1,36 @@
...
-    </div>
-  </div>
-<% end %>
-
   <div class="form-group col-xs-12">
     <div class="col-xs-offset-2 col-xs-10">
-      <%= f.submit class: "btn btn-primary", data: { confirm: 'Are you sure?' } %>
-      <%= link_to 'Back', :back, class: "btn btn-default" %>
+      <%= f.submit class: "btn btn-primary", data: { confirm: "Are you sure?" } %>
+      <%= link_to "Back", :back, class: "btn btn-default" %>
```
Anbranin: My only gripe so far with your changes is this "I need double-quotes" thing, but it's not a blocking concern because it really doesn't matter that much. That being said, it wasn't worth putting effort into. You don't need double-quotes in ruby unless you a) have a single quote within it or b) are doing interpolation of some kind.

app/views/incurred_incidentals/_form.html.erb
```diff
@@ -1,54 +1,66 @@
 <%= form_for setup_incurred_incidental(@incurred_incidental), html: { multipart: true } do |f| %>
   <div>
-    <%= f.hidden_field :rental_id, value: params[:rental_id] || @incurred_incidental.rental.id, class: "selectpicker form-control" %>
+    <%= f.hidden_field :rental_id, value: params[:rental_id] || @incurred_incidental.rental.id,
+                                   class: "selectpicker form-control" %>
   </div>
 
   <div class="form-group col-xs-12">
-    <%= f.label :item_id, 'Item', class: 'control-label col-xs-2' %>
+    <%= f.label :item_id, "Item", class: "control-label col-xs-2" %>
```
Anbranin: this is killing me

app/views/incurred_incidentals/_form.html.erb
```diff
@@ -1,54 +1,66 @@
...
       <% end %>
     <% end %>
   </div>
 
   <div class="form-group col-xs-12">
-    <%= label_tag :documents, 'Documents', class: 'control-label col-xs-2' %>
+    <%= label_tag :documents, "Documents", class: "control-label col-xs-2" %>
     <div class="col-xs-6">
-      <%= render partial: 'documents/upload', locals: { parent: f } %>
+      <%= render partial: "documents/upload", locals: { parent: f } %>
```
Anbranin: 😿

app/views/rentals/_form.html.erb
```diff
@@ -2,43 +2,57 @@
...
 
       <div class="form-group row">
-        <%= f.label :start_time, 'Pickup Date', class: 'control-label col-xs-3' %>
+        <%= f.label :start_time, "Pickup Date", class: "control-label col-xs-3" %>
         <div class="col-xs-9">
-          <%= f.date_field :start_time, value: @start_date, required: true, id: 'rental_start_time', class: "form-control cost-dependent" %>
+          <%= f.date_field :start_time, value: @start_date,
+                                        id: "rental_start_time",
+                                        class: "form-control cost-dependent",
+                                        required: true %>
```
Anbranin: I like that you decreased line length though

app/views/incidental_types/_form.html.erb
anmerkl: While I almost always prefer single quotes in code, I like to almost exclusively use double quotes in html. In shared code-bases where everybody has their own style, I'll change everything to double quotes for consistency. It wasn't any effort at all, it was a simple find-replace all (and make sure it doesn't break any embedded quotes situation)

anmerkl:
Completely forgot to run the specs
Please hold off on reviewing until I can fix the issues

app/views/incidental_types/_form.html.erb
Anbranin: Fine, but in that case I'll point out a couple things: 
One, it added a lot of bulk to a PR that was already pretty bulky, and wasn't a functional change. I'd have put it in its own PR.
Two, this isn't HTML in this context--while the document gets turned into HTML, every place you've used double quotes here is Ruby (and necessarily so, because if it were HTML, they'd get read as plain text).
Three: As an organization, we (okay, at least here at Transit, Parking was pretty separate for a while) prefer single quotes unless double-quotes are necessary--because it's easier to find interpolation when scanning a document that way.
Either way, any argument for or against single vs double quotes is pretty weak, so this is not a hill I'm going to die on, do what you like. Haha

app/models/user.rb
```diff
@@ -2,18 +2,24 @@
...
-
-  validates :first_name, :last_name, :spire_id, :phone, :email, :department, presence: true
-  validates :spire_id, length: { is: 8 }, uniqueness: true
+  has_many :rented_rentals, class_name: Rental.name, foreign_key: :renter_id
+
+  validates :first_name, :last_name, :spire_id,
+            :department, :phone, :email, presence: true
+  validates :spire_id, uniqueness: true,
+             format: { with: /\d{8}/, message: 'SpireId should be 8 digits' }
+  validates :phone, format: { with: /\d{10}/,
```
Cobchise: I wonder if here it would be better to account for different formats of phone numbers instead of just checking for 10 digits (which also only accounts for US numbers.. I'm not sure if we'd ever get an international number on this app but currently the app wouldn't allow it).
What about something like
`validates :phone, format: {
      :with => /\(?[0-9]{3}\)?-[0-9]{3}-[0-9]{4}/,
      :message => "- Phone numbers must be in xxx-xxx-xxxx format."
}` which looks for xxx-xxx-xxxx format but also allows optional parenthesis. (Again still doesn't account for non-US numbers)

app/models/payment.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 class Payment < ApplicationRecord
-  has_one :financial_transaction, as: :transactable
   enum payment_type: [:recharge]
-  validates :payment_type, :contact_name, :contact_email, :contact_phone, presence: true
-  validates :contact_phone, length: { is: 10 }
+  has_one :financial_transaction, as: :transactable
+  validates :payment_type, :contact_name, :contact_phone, :contact_email, presence: true
+  validates :contact_phone, format: { with: /\d{10}/,
```
Cobchise: Same here

app/assets/javascripts/financial_transactions/new.js
```diff
@@ -1,17 +1,17 @@
...
     if(!email_regex.test($('#contact_email').val()))
       $('#contact_email').parent().parent().addClass('has-error');
-    else 
+    else
       $('#contact_email').parent().parent().removeClass('has-error');
   });
 
   $('#contact_phone').change(function (){
     var email_regex = /^[0-9]{10}$/i;
```
Cobchise: These vars should be 'phone_regex' right? Also I wonder if it would be better to look for formatted phone numbers here and not just a string of 10 digits. Perhaps a regex like the following:
`/\(?[0-9]{3}\)?-[0-9]{3}-[0-9]{4}/` which looks for xxx-xxx-xxxx format but also allows optional parenthesis (if we allow optional parenthesis we should add a note to the field or to the placeholders alerting the user they can user parenthesis if they wish). Jquery validator may also be able to handle this as @Anbranin mentioned. And as I said in another comment, perhaps allow more than 10 digits to account for non-US numbers also

app/models/user.rb
Anbranin: I agree with that; we should allow for international numbers. It's an edge case, but we are a fairly international campus here.
I think validating that the phone number is a digit before saving is sufficient.
I would also allow entering the phone number in any way the person sees fit and then stripping any non-numeric characters before saving. For example, we should allow someone to type in `413-545-0056` or `1+413-545-0056` or `(413)545-0056`, which are all a) valid ways to display or type a phone number and b) allowed on practically any other website (which isn't shite) that asks for a number. Then use a `before_save` block to strip any non alphanumeric characters.
If a person enters in a letter, perhaps we should send an error, because it probably means they sausage-fingered and should check the number.

app/views/financial_transactions/_form.html.erb
```diff
@@ -16,52 +16,55 @@
...
+      <%= text_field_tag :contact_email, nil, class: "form-control", required: true %>
     </div>
   </div>
 
   <div class="form-group col-xs-12">
-    <%= label_tag :contact_phone, 'Contact Phone', class: 'control-label col-xs-2' %>
+    <%= label_tag :contact_phone, "Contact Phone", class: "control-label col-xs-2" %>
     <div class="col-xs-6">
-      <%= text_field_tag :contact_phone, nil, class: 'form-control', placeholder: '5555555555', required: true %>
+      <%= text_field_tag :contact_phone, nil, placeholder: "5555555555",
```
Cobchise: This placeholder is not consistent with the user form which displays a placeholder of 555-555-5555. I would like to see them perhaps both change to XXX-XXX-XXXX

app/views/users/_form.html.erb
```diff
@@ -35,14 +41,14 @@
   <div class="form-group col-xs-12">
     <label class="col-xs-2 control-label" for="phone">Phone Number</label>
     <div class="col-xs-6">
-      <%= text_field :user, :phone, size: 20, placeholder: "111-222-3333", class:"form-control", required: true %>
+      <%= text_field :user, :phone, placeholder: "111-222-3333", class:"form-control", required: true %>
```
Cobchise: This is the line I was referring to above, again would like to see this display XXX-XXX-XXXX

app/views/financial_transactions/_form.html.erb
Anbranin: One could also use a mask. I'm sure there's a plugin available. But yeah, at least not a string of digits like this.

app/assets/javascripts/financial_transactions/new.js
anmerkl: I'm on board with all of this.
I'm just going to scrap this function.  Currently, all it does is give the input box a red outline if the input doesn't match this regex. It doesn't actually stop the user from submitting even if the format doesn't match.  Since I'm going to make a change to allow for various format inputs, the regex would be unnecessarily complicated.

spec/models/payment_spec.rb
```diff
@@ -23,8 +23,8 @@
       expect(build(:payment, contact_phone: nil)).not_to be_valid
     end
 
-    it 'requires length contact_phone to be 10' do
-      expect(build(:payment, contact_phone: '1' * 9)).not_to be_valid
+    it 'requires length contact_phone to be 8' do
```
Anbranin: We allow numbers _over_ 8, right?

anmerkl: Yes

anmerkl:
Ready for re-review

.codeclimate.yml
```diff
@@ -30,15 +30,20 @@ engines:
     enabled: true
   fixme:
     enabled: true
+    checks:
+      XXX:
+        enabled: false
```
anmerkl: For some reason Rubocop doesn't like the 'XXX-XXX-XXXX' placeholders that we discussed so I'm disabling that check.

.codeclimate.yml
```diff
@@ -30,15 +30,20 @@ engines:
...
     enabled: true
     checks:
       Rubocop/Style/SingleLineBlockParams:
         enabled: false
-      Rubocop/Rails/DynamicFindBy:
-        enabled: false
       Rubocop/Style/AndOr:
         enabled: false
+      Rubocop/Style/IfUnlessModifier:
+        enabled: false
```
anmerkl: This cop enforces that if `if` and `unless` blocks are only a single line the conditional should be moved onto the same line instead of using a block.  However, this often causes the resulting line to be unreasonable long.  If I had to choose between the two, I would choose manageable and reasonable lines so I'm disabling this cop.

anmerkl:
This is FINALLY ready to go - the last issue I've been tackling all week was introduced because I reordered some lines of code...

Issue #382, opened 2018-02-05
Clean up errors


Cobchise:
We need to clean up errors across the board. Errors need to be succinct but also provide some indication as to what exactly went wrong. 

An example of this is that the inventory API has a handle errors call before returning during delete_reservations on rentals, but instead of using those errors, we dismiss them and force a generic error instead. This is poor practice, we want our errors to be helpful and specific to the problem they are referencing.

anmerkl:
I'm assuming I'm only improving on the fact that all RuntimeErrors are generically caught and handled by this line:
`rescue_from RuntimeError, Exception, with: :render_500`

Are there any other errors that need improvements? 404? Flash errors?

Anbranin:
Yeah rescuing from everything in the application controller is not the best idea
https://github.com/umts/umaps-rails/pull/819

Cobchise:
Flash errors definitely need improvement, they're actually the main reason I created this issue. Most of the flash errors in the app just really aren't helpful to the user. 

A lot of the 'error-able' methods in the app actually do have detailed flash messages returning, however at some point they are abstracted or overwritten by a parent method which just dumps non-readable and non-specific rails errors to the user instead of a the original nice message such as 'Creating the rental failed because there are no available carts'.

PR #398, opened 2018-03-01
Financial transactions fix

### LINKED ISSUE #392: Financial Transactions controller/view issues
- Filter function on the index page doesn't filter anything
- show view only has edit and back buttons
- edit form does not repopulate the fields
- edit has no corresponding update method in the controller

anmerkl:
closes #392 

Fixed filter on index page (though with a workaround)
Removed unneeded actions and views

### DIFF
--- a/app/controllers/financial_transactions_controller.rb
+++ b/app/controllers/financial_transactions_controller.rb
@@ -4,12 +4,7 @@ class FinancialTransactionsController < ApplicationController
 
   def index
     # workaround for unknown Ransack bug
-    @q = if params[:q]
-           FinancialTransaction.search(transactable_type_eq: params[:q][:transactable_type])
-         else
-           FinancialTransaction.search(nil)
-         end
-
+    @q = FinancialTransaction.search(transactable_type_eq: params[:q].try(:[], :transactable_type))
     @trans_type = FinancialTransaction.all.pluck(:transactable_type).uniq
     @financial_transactions = @q.result.paginate(page: params[:page], per_page: 10)
   end
--- a/app/controllers/financial_transactions_controller.rb
+++ b/app/controllers/financial_transactions_controller.rb
@@ -8,8 +8,8 @@ class FinancialTransactionsController < ApplicationController
            FinancialTransaction.search(transactable_type_eq: params[:q][:transactable_type])
          else
            FinancialTransaction.search(nil)
-    end
-    
+         end
+
     @trans_type = FinancialTransaction.all.pluck(:transactable_type).uniq
     @financial_transactions = @q.result.paginate(page: params[:page], per_page: 10)
   end
@@ -20,9 +20,7 @@ class FinancialTransactionsController < ApplicationController
     @financial_transaction.transactable_type = params[:transactable_type]
 
     # handles transactable_type payment which will be created with this form
-    if @financial_transaction.transactable_type != Payment.name
-      @financial_transaction.transactable_id = params[:transactable_id]
-    end
+    @financial_transaction.transactable_id = params[:transactable_id] if @financial_transaction.transactable_type != Payment.name
   end
 
   def create
--- a/app/controllers/financial_transactions_controller.rb
+++ b/app/controllers/financial_transactions_controller.rb
@@ -4,15 +4,14 @@ class FinancialTransactionsController < ApplicationController
 
   after_action :set_return_url, only: [:index]
 
-  # GET /financial_transactions
+  def show
+  end
+
   def index
     @q = FinancialTransaction.search(params[:q])
     @trans_type = FinancialTransaction.all.pluck(:transactable_type).uniq
     @financial_transactions = @q.result.paginate(page: params[:page], per_page: 10)
   end
-
-  # GET /financial_transactions/1
-  def show; end
 
   def new
     @financial_transaction = FinancialTransaction.new
@@ -23,10 +22,6 @@ class FinancialTransactionsController < ApplicationController
     @financial_transaction.transactable_id = params[:transactable_id] if @financial_transaction.transactable_type != Payment.name
   end
 
-  # GET /financial_transactions/1/edit
-  def edit; end
-
-  # POST /financial_transactions
   def create
     @financial_transaction = FinancialTransaction.new(financial_transaction_params)
     if @financial_transaction.transactable_type == Payment.name
@@ -45,19 +40,23 @@ class FinancialTransactionsController < ApplicationController
     end
   end
 
+  def edit
+  end
+
+  def update
+  end
+
   private
 
-  # Use callbacks to share common setup or constraints between actions.
   def set_financial_transaction
     @financial_transaction = FinancialTransaction.find(params[:id])
   end
 
-  def payment_params
-    params.permit(:payment_type, :contact_name, :contact_email, :contact_phone, :reference)
-  end
-
-  # Only allow a trusted parameter "white list" through.
   def financial_transaction_params
     params.require(:financial_transaction).permit(:amount, :rental_id, :transactable_type, :transactable_id)
+  end
+
+  def payment_params
+    params.permit(:payment_type, :contact_name, :contact_email, :contact_phone, :reference)
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/financial_transactions_controller.rb
```diff
@@ -1,63 +1,56 @@
 # frozen_string_literal: true
 class FinancialTransactionsController < ApplicationController
-  before_action :set_financial_transaction, only: [:show, :edit]
-
   after_action :set_return_url, only: [:index]
 
-  # GET /financial_transactions
   def index
-    @q = FinancialTransaction.search(params[:q])
+    # workaround for unknown Ransack bug
```
anmerkl: The regular Ransack syntax `FinancialTransaction.search(params[:q])` was not working, which is why filter on the index page wasn't working.  For some reason, Ransack was not actually searching for anything as if `params[:q]` was `nil`.

config/routes.rb
```diff
@@ -43,7 +41,7 @@
   resources :holds do
     post :lift, on: :member
   end
-  resources :financial_transaction, except: %i(destroy update)
```
anmerkl: I'm basing changes on this line as opposed to the above line because Metin said he did not want Financial Transactions to be deleted or edited after creation

app/controllers/financial_transactions_controller.rb
```diff
@@ -23,41 +23,33 @@ def new
...
-  def edit; end
-
-  # POST /financial_transactions
   def create
     @financial_transaction = FinancialTransaction.new(financial_transaction_params)
     if @financial_transaction.transactable_type == Payment.name
-      payment = Payment.new(payment_params) # hard fail
+      payment = Payment.new(payment_params)
       unless payment.save
         flash[:danger] = 'Please Properly Fill Out Contact And Payment Fields'
```
Anbranin: Would it be better to be clearer here, like so:
`flash[:danger] = payment.errors.full_messages`?
That is, if your application.html.erb is set up to accept flash[:danger] as an array.

Anbranin: If that's outside of the scope of your PR, never mind--I just think it would be good to do.

app/controllers/financial_transactions_controller.rb
```diff
@@ -23,41 +23,33 @@ def new
...
+        render :new and return
       end
+
       @financial_transaction.transactable_id = payment.id
     end
 
     if @financial_transaction.save
-      redirect_to rental_invoice_path(@financial_transaction.rental_id), success: 'Financial Transaction Successfully Created'
+      redirect_to rental_invoice_path(@financial_transaction.rental_id),
+                  success: 'Financial Transaction Successfully Created'
```
Anbranin: Shouldn't this be in the flash?

app/controllers/financial_transactions_controller.rb
anmerkl: I agree, but I chose not to address that here because it will be addressed when I eventually finish #382

app/controllers/financial_transactions_controller.rb
Cobchise: Ransack usually doesn't return anything when params[:q] is nil, that's correct behavior as far as I can tell. Some people choose to apply their own filter when params[:q] is nil like below:
```
def index
  @search = Job.search(params[:q])
  @results = @search.result
  @results = @results.where(:your_date => nil) unless params[:q]
  @job = @results.paginate(:per_page => 10, :page => params[:page])
end
``` 
but I found a better example which actually changes the default return of the search object which may be a better approach:
```
def index
  @search = Job.search(params[:q])
  @search.status_cont = 'Open' unless params[:q]
  @results = @search.result
  @job = @results.paginate(:per_page => 10, :page => params[:page])
end
```
Hope this helps. Both of these can be seen [here](https://stackoverflow.com/questions/13284172/setting-default-search-parameter-on-ransack-for-rails)

spec/controllers/financial_transactions_controller_spec.rb
```diff
@@ -138,22 +122,4 @@
...
-  describe 'PUT #update' do
-    it 'cannot find a route' do
-      ft = create :financial_transaction
-      expect do
-        put :update, params: { id: ft.to_param }
-      end.to raise_error AbstractController::ActionNotFound
-    end
-  end
-
-  describe 'DELETE #destroy' do
```
Cobchise: I don't think you need to remove the destroy method as our permanent_records gem overrides the behavior. Instead of deleting the record when you call .destroy it adds a 'deleted_at' datetime column entry.

config/routes.rb
Cobchise: Based again on the permanent_records gem I believe we can allow the destroy resource also, as permanent_records (as I noted below) overrides the destroy behavior

app/controllers/financial_transactions_controller.rb
anmerkl: The issue here was not that `params[:q]` was nil.  It was behaving as though `params[:q]` was nil.
If `params[:q]` is nil, Ransack doesn't do any filtering so it returns everything, which is why the filter function on the index page wasn't working.

app/controllers/financial_transactions_controller.rb
anmerkl: Yes, that will also be addressed in #382

spec/controllers/financial_transactions_controller_spec.rb
anmerkl: We actually never had a destroy method so I figured we didn't want it in the first place.

Cobchise: Apologies, yep you're correct to remove this if the method it is testing doesn't exist... but it just begs the question why was it here at all? I'm still wondering why we don't have a destroy method for financial transactions

anmerkl: I'd like to think this ended up here after a bunch of copy-pasting + making relevant edits to make the tests pass

We don't delete Financial Transactions because we'd like to keep track of everything related to money so it doesn't make sense to just delete these records.  We could consider adding a functionality to 'deactivate' Financial Transactions, but I don't think that that adds any useful functionality.

app/controllers/financial_transactions_controller.rb
anmerkl: Also, 
```
if params[:q]
  FinancialTransaction.search(transactable_type_eq: params[:q][:transactable_type])
...
```
looks to be logically equivalent to what this snippet is accomplishing
```
@results = @search.result
@results = @results.where(:your_date => nil) unless params[:q]
```
So there shouldn't be a problem.

spec/controllers/financial_transactions_controller_spec.rb
Cobchise: Well that was my point above, we already have permanent_records implemented so if we made a destroy method they wouldn't be deleted as permanent_records overrides that functionality. Making a destroy method would be effectively making a 'deactivate' method.

app/controllers/financial_transactions_controller.rb
Cobchise: It largely is, but I suggested it an an improvement because it doesn't require a weird else statement where you pass `nil` to FinancialTransaction.search

spec/controllers/financial_transactions_controller_spec.rb
anmerkl: I see.
I'd rather put that in a new issue and pull request then, based on what I read from the permanent_records documentation.  I think what is currently in this pull request is enough for the MVP. What do you think?

Cobchise: Agreed. I'll make the issue now

app/controllers/financial_transactions_controller.rb
anmerkl: What do you think about the below implementation? It's a bit more complicated but is a lot cleaner

Cobchise: Nice code man, I like that a lot. Thanks for changing it :+1: As long as it works then lgtm

PR #386, opened 2018-02-12
Gemfile update and Rails update

Cobchise:
### Closes #384 

Removed some version numbers from gemfile and ran bundle update. Also updated rails to 5.1.4

PR #399, opened 2018-03-01
Help page

elisacoutant:
Puts an Asterix next to all the required fields on forms.
Also populated the Help page with some basic steps for the most important form.


Anbranin:
@elisacoutant, Putting the asterisks in automatically with js/css can be done the same way as it's done in UMaps-Rails, which I think is fine. See [here](https://github.com/umts/umaps-rails/blob/master/app/assets/javascripts/ui.js#L20-L26) and [here](https://github.com/umts/umaps-rails/blob/master/app/assets/stylesheets/forms.css#L21-L35).

app/views/groups/_form.html.erb
```diff
@@ -1,27 +1,27 @@
 <%= form_for(@group) do |f| %>
   <div class="form-group col-xs-12">
-    <label class="col-xs-2 control-label" for="name">Name</label>
+    <label class="col-xs-2 control-label" for="name">Name*</label>
```
Cobchise: Take out the *'s next the the field names in each of these

app/controllers/help_controller.rb
```diff
@@ -0,0 +1,4 @@
+class HelpController < ApplicationController
+    def show
+    end
+  end
```
Cobchise: Indentation is a bit weird here

public/ui.js
```diff
@@ -0,0 +1,7 @@
+/* Adds a red asterix next to required fields */
+$(function(){
```
Cobchise: I feel as though this code should be in app/assets/javascripts/application.js or a relevant file at the same path like probable-engine/app/assets/javascripts/forms.js
That way rails renders it into the public folder through pre-compilation in production.

app/views/users/_form.html.erb
```diff
@@ -1,29 +1,29 @@
...
         <%= text_field :user, :spire_id, size: 20, placeholder: "1234567", class:"form-control", required: true %>
       </div>
     </div>
   <% end %>
 
   <div class="form-group col-xs-12">
-    <label class="col-xs-2 control-label" for="department_id">Department</label>
+    <label class="col-xs-2 control-label" for="department_id">Department*</label>
     <div class="input-group col-xs-6">
       <%= f.select :department_id, options_from_collection_for_select(@departments, "id", "name", @user.department_id), {}, class: "selectpicker form-control", onchange: "update_disclaimer(this)", required: true %>
```
Anbranin: Can we get line length under control a little here? I'm not going to be a stickler for exactly 80 characters but this is insane.

elisacoutant:
Help page now has explanations for each of the 8 forms; Metin approved.

Anbranin:
Great PR seems like you guys have it under control but for the love of the English language it is spelled _asterisk_

Gemfile.lock
```diff
@@ -9,7 +9,8 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    aasm (4.11.1)
+    aasm (4.12.3)
+      concurrent-ruby (~> 1.0)
```
anmerkl: Where are these changes coming from?

app/views/help/new_department.html.erb
```diff
@@ -0,0 +1,10 @@
+<% provide :page_header, 'Help-New Department' %>
+
+<body>
+  <h4>New Department Form </h4>
+    <ol>
+      <li>Type in a 'Department Name.'</li>
+      <li>Click on 'Create Deparment' and wait for success message.</li>
```
Cobchise: missing 't' in 'Deparment'

bin/spring
```diff
@@ -7,9 +7,11 @@ unless defined?(Spring)
   require 'rubygems'
   require 'bundler'
 
-  if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
-    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq }
-    gem 'spring', match[1]
+  lockfile = Bundler::LockfileParser.new(Bundler.default_lockfile.read)
```
Cobchise: What's this?

anmerkl: Just tried to pull down your branch to view the help pages.
Trying to run `bundle install` returns this error:
```
Your bundle is locked to parser (2.5.0.4), but that version could not be found in any of the sources listed in your Gemfile.
If you haven't changed sources, that means the author of parser (2.5.0.4) has removed it.
You'll need to update your bundle to a different version of parser (2.5.0.4) that hasn't been removed in order to install.
```

elisacoutant: Not sure what happened here but its resolved in my latest commit.

Gemfile.lock
anmerkl: I was getting weird dependency errors because of changes in this file.
This is a little suspect because there's no equivalent changes to the `Gemfile`.  Since we're already handling this in #386, I'd like for these changes to be reverted.

bin/spring
anmerkl: Kind of related to my above comment, but if this is was an unintentional change then we shouldn't merge it into master.  I just reverted this file and the `Gemfile.lock` and everything still runs fine.
This should be reverted as well.

app/assets/javascripts/application.js
```diff
@@ -34,3 +34,11 @@ $(document).ready(function() {
 });
 
 $( ".datepicker" ).datetimepicker(); //datepicker initialization
+
+/* Adds a red asterix next to required fields */
```
Anbranin: asterisk. ;_; plz

app/controllers/help_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class HelpController < ApplicationController
+  def show; end
```
Anbranin: Do you need empty controller actions for pages in rails 5+?

config/routes.rb
```diff
@@ -9,8 +9,19 @@
   get 'rentals/:id/invoice', to: 'rentals#invoice', as: 'rental_invoice'
   get 'rentals/safety_pdf'
   get 'rentals/training_pdf'
-  resources :rentals
 
+  #---Help Forms-------------------------------------------------------------
+  get 'help', to: 'help#show'
```
Anbranin: Curious why you're constructing your routes this way. Do you need all these individual path helpers? (`as: 'path'`)
What if you did 
``` ruby
resources :help, :none do
  collection do  
    get :schedule
    etc
  end
end
```
I think you'd have the exact same path helpers in most cases, also, and it would be a lot cleaner.

app/controllers/help_controller.rb
anmerkl: Doesn't look like they're needed.
http://guides.rubyonrails.org/layouts_and_rendering.html#rendering-by-default-convention-over-configuration-in-action

config/routes.rb
elisacoutant: True, will change

config/routes.rb
```diff
@@ -9,8 +9,22 @@
   get 'rentals/:id/invoice', to: 'rentals#invoice', as: 'rental_invoice'
   get 'rentals/safety_pdf'
   get 'rentals/training_pdf'
-  resources :rentals
```
Anbranin: Aren't there resourceful routes for the rentals?

nawilliams: This line is still there, just below the help routes. I guess it got inserted weird, I'll move it back.

Anbranin: Ah alright I just missed that fact.

PR #402, opened 2018-03-27
Show available rentals

anmerkl:
Add table to Rentals form that displays the number of available items for each item type

### DIFF
--- a/app/controllers/rentals_controller.rb
+++ b/app/controllers/rentals_controller.rb
@@ -26,7 +26,7 @@ class RentalsController < ApplicationController
     # this probably isnt the most efficient way to do it
 
     @q = base_search_area.search(params[:q])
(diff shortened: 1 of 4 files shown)


anmerkl:
screen shot 2018-03-27 at 09 46 25

anmerkl:
My only gripe with this currently is that there doesn't seem to be a way to make it so that the user sees available items specifically for the date range they have selected.
My original plan was to get a list of reservations and cross-reference these reservation dates with the user-selected dates to get an accurate number of available items based on the user-selected dates.  However, the end point on `aggressive-epsilon` for getting reservations for a date range returns a JSON of ONLY the start and end dates of the reservation - there is no indication of which item the reservation dates belong to.
Unless someone knows a better way to do this, I'm throwing the code as is up for MVP.  The improvement that I wanted to implement would depend on first making an update to `aggressive-epsilon` to get a more informative list of reservations.

Issue #393, opened 2018-02-23
search/filter-related elements are clunky


anmerkl:
Search and filter related elements (buttons, dropdowns, etc.) look cramped.
For example, `/rentals/processing` has dropdowns that allow a user to specify items for search.  The dropdowns then expand to include every item that is specified by the user, causing every element following the dropdown to have to shift farther right or down one line, leaving for a clunky UI.

This should be cleaned up at some point.

anmerkl:
closed with #403

PR #403, opened 2018-03-27
View cleanup

elisacoutant:
Resolves Issue #393 

In the rentals processing/processing page.
I changed the search button to fit in with the buttons on the other forms.

I also spaced out the form so the pickers weren't on top of each other and now they each have their own line. Open to suggestions on how to do it a different way, I just thought it looked good this way.

### DIFF
--- a/app/views/rentals/processing.html.erb
+++ b/app/views/rentals/processing.html.erb
@@ -3,15 +3,15 @@
 <%= search_form_for @q, url: '/rentals/processing', class: 'form-inline' do |f| %>
   <%= label_tag :item_type_id_in, 'Containing Item Types' %>
   <%= select_tag :item_type_id_in, options_from_collection_for_select(@item_types.order(name: :asc), "id", "name", selected: params[:item_type_id_in]),
-    multiple: true, include_blank: true, class: "selectpicker form-control col-xs-2" %>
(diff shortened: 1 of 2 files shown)


app/views/rentals/processing.html.erb
```diff
@@ -3,23 +3,23 @@
 <%= search_form_for @q, url: '/rentals/processing', class: 'form-inline' do |f| %>
   <%= label_tag :item_type_id_in, 'Containing Item Types' %>
   <%= select_tag :item_type_id_in, options_from_collection_for_select(@item_types.order(name: :asc), "id", "name", selected: params[:item_type_id_in]),
-    multiple: true, include_blank: true, class: "selectpicker form-control col-xs-2" %>
+    multiple: true, include_blank: true, class: "selectpicker form-control col-xs-2" %> <br></br>
```
anmerkl: Just a note:
`<br>` is one of those tags that doesn't to be explicitly closed because nothing is even allowed to be between the opening and closing tag.
You could replace this with just a `<br>` by itself without the closing tag or if you want to be a bit more concise you can also use `<br />`

elisacoutant: I tried those originally, but the space between the rows still looked clunky.

elisacoutant:
After going through I found a similar issue under rentals/index and fixed it to match the other one. 
I don't see any other pages that have this issue.

PR #405, opened 2018-04-17
Error cleanup

### LINKED ISSUE #382: Clean up errors

anmerkl:
resolves #382 

Edited `flash` messages so that they actually show up
Added new error handler for `InventoryErrors` so that they are no longer handled as generic 500 errors.


app/controllers/damages_controller.rb
```diff
@@ -30,8 +30,7 @@ def create
     end
   end
 
-  def edit
-  end
+  def edit; end
```
Anbranin: Do you need an empty controller action?

app/controllers/departments_controller.rb
```diff
@@ -8,8 +8,7 @@ def index
     @departments = Department.all
   end
 
-  def show
-  end
+  def show; end
```
Anbranin: Same question here.

app/controllers/damages_controller.rb
anmerkl: Forgot you mentioned that in another PR and that I confirmed it...

Anbranin:
Ignore my comments about needing empty controller actions--you don't normally need them but I realized the permissions system probably depends on them existing.

anmerkl:
Oh... that makes sense...

.codeclimate.yml
```diff
@@ -42,6 +42,8 @@ engines:
         enabled: false
       Rubocop/Style/IfUnlessModifier:
         enabled: false
+      Rubocop/Style/GuardClause:
+        enabled: false
```
anmerkl: Disabling this for the same reason I disabled the one line if/unless statements - rubocop yells at us for trying to pragmatically keep line length down

app/controllers/damages_controller.rb
```diff
@@ -11,31 +11,35 @@ def index
...
 
     if @damage.save
-      flash[:success] = 'Damage Successfully Created'
+      flash[:success] = 'Damage successfully created'
       redirect_to @damage
     else
-      flash[:warning] = 'Failed To Create Damage'
-      render 'new'
+      flash[:danger] = 'Failed to create Damage'
+      flash[:warning] = @damage.errors.full_messages
```
Anbranin: So, as a user if I see two flash boxes of different colors, I would assume that there are two things wrong, and also may not assume that the things are even connected. Since there's just one thing wrong, we should use just one flash message box.

anmerkl: Ok, I'll remove the `flash[:danger]` messages

Anbranin: There's no reason you can't have both messages in there if you'd like to, although it's not something I'd think of requesting. `@damage.errors.full_messages` is an array, so the layout must have logic to loop through arrays and display each error in turn. Therefore you could do this:
``` ruby
flash[:danger] = 'Failed to create damage'
flash[:danger] << @damage.errors.full_messages
```
As for which flash message to use, I'd use whichever is red. If something fails to save, it's a big bad and people should know that the thing they just did failed in no uncertain terms. Yellow is for when something might be slightly off, and green is success... at least that's how I've always seen it used. 
I'd interpret "warning" to be "hey, something worked, but also something didn't or may not and you should know about it."

Anbranin: Also, if you do go with putting the first message in the flash, there may be a better way than manually putting it there upon every failed request in the controller. Like in the model.

anmerkl: Oops, did not see these comments.

The layout does have logic to loop over arrays which is why I changed things like
`@department.errors.full_messages.each { |e| flash_message :warning, e, :now }`
to
`flash[:danger] = @department.errors.full_messages`

I actually tried something like your example before, but it doesn't work because of a String to Array conversion error.  I then tried doing something like
```
flash[:danger] = ['Failed to create damage:']
flash[:danger] << @damage.errors.full_messages
```
to circumvent that issue, but that results in oddly formatted error messages, because it'd require having a double loop in the layout, which I didn't think was necessary.

I don't think having both messages is necessary after making the change.  I will change the errors to `:danger` because you're right in that red should signify that something went wrong, which I think is also enough to imply to the user that creating or updating something failed.

Anbranin: Oh man you're right. I remembered it having been done before but I was remembering shoveling a string into an existing array, not the thing I said which is syntactically incorrect. Carry on, I think you're right, the baked-in error messages are traditionally enough.

spec/controllers/items_controller_spec.rb
```diff
@@ -102,8 +102,8 @@
     context 'with no name' do
       it 'populates a danger flash message' do
         post :create_item, params: { name: nil, type: item_type.name }
-        expect(flash[:warning]).to be_present
-        expect(flash[:warning]).to eq('Name is a required field')
+        expect(flash[:danger]).to be_present
```
Anbranin: Not a blocking concern, I just want to mention that checking that `flash[:danger]` is eql to a string is definitely also checking that it's present.

PR #407, opened 2018-04-23
Improve back buttons

### LINKED ISSUE #395: Unexpected Back button behavior
The Back button should ideally redirect to the last page the user was on.  However, if a user lands on the same page more than once (for example, due to incorrect form submission), the Back button will simply reload that same page over and over again.

While I guess that kind of makes sense logically, it's not a good user experience.
Haven't checked every single Back button, but one example is on the `incurred_incidentals` form. And if it's happening in one place, I'm betting it's also happening somewhere else.

anmerkl:
resolves #395 

Back buttons on forms would lead the user in a loop if an error occurs upon submitting the form.
Made use of existing `link_back` and `set_return_url` to make back button behavior more consistent.

### DIFF
--- a/app/views/damages/_form.html.erb
+++ b/app/views/damages/_form.html.erb
@@ -65,7 +65,6 @@
   <div class='form-group col-xs-12'>
     <div class='col-xs-offset-2 col-xs-10'>
       <%= f.submit class: 'btn btn-primary', data: { confirm: 'Are you sure?' } %>
-      <%= link_back 'Back', damages_path, class: 'btn btn-default' %>
     </div>
   </div>
 <% end %>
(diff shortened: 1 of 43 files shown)


app/views/damages/_form.html.erb
```diff
@@ -65,7 +65,7 @@
   <div class='form-group col-xs-12'>
     <div class='col-xs-offset-2 col-xs-10'>
       <%= f.submit class: 'btn btn-primary', data: { confirm: 'Are you sure?' } %>
-      <%= link_to 'Back', :back, class: 'btn btn-default' %>
+      <%= link_back 'Back', damages_path, class: 'btn btn-default' %>
```
Anbranin: What problem is this solving? Why isn't `link_to 'Back', :back` sufficient?

anmerkl:
The changes in this PR are no longer relevant, since I'm removing back buttons entirely

PR #408, opened 2018-04-24
Back buttons

### LINKED ISSUE #395: Unexpected Back button behavior

anmerkl:
resolves #395 

After a discussion with Karen, having back buttons on this site isn't really useful since a user's browser already has a back button.  Any issues concerning our back buttons could be circumvented using the browser's back button, so I'm just taking back buttons out entirely.
Adding a 'home' button was mentioned to allow a user to return to the default page from anywhere, but that already exists in the navigation bar.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -117,9 +117,4 @@ class ApplicationController < ActionController::Base
       (params[:controller] == 'home' && params[:action] == 'index') ||
       params[:action] == 'render_404'
   end
-
-  def set_return_url
-    # save the current url for use by back buttons
-    session[:return_url] = request.original_url
-  end
 end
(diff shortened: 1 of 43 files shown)


Anbranin:
*Karin

Anbranin:
I should have waited for @Cobchise's review--I'll defer to his judgement, since you can still review closed pull requests I'll revert if he has issues.

Cobchise:
No worries, and no issues from me. Good PR 👍

PR #409, opened 2018-04-26
Add system tests

Anbranin:
Added the capability to write system tests. ~To run them, you must have chromedriver installed on your mac. You can use brew cask for this~. Update: The gem `chromedriver-helper` should install chromedriver on your machine.
~I wrote one example system test that definitely fails on my machine, but only because I can't actually get this app to run for me.~ I had to remove it since it fails on travis. The travis configuration file needs to be updated to somehow run aggressive-whatever-it-is, which I don't feel like doing.
So basically,  this PR upgrades to rails 5.1 and sets it up so that system tests _can_ be written. (I had to upgrade the rails version so that system testing would work at all).


spec/system/rental_creation_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+describe 'creating a new rental' do
+  before :each do
+    @tomorrow = Time.zone.tomorrow
+    @day_after = @tomorrow + 1.day
+  end
+
+  describe 'form autopopulation' do
+    before :each do
+      FactoryBot.create :user
+      visit '/rentals/new'
```
nawilliams: You should just be able to call create :user as the FactoryBot methods are included in the rails_helper.

app/views/layouts/shared/_loading.html.erb
```diff
@@ -1,6 +1,6 @@
 <div id="loading">
 	<div id="loading-text">
 		Loading<br/>
-		<%= image_tag("loading.gif") %>
+		<%= image_tag"loading.gif", skip_pipeline: true %>
```
nawilliams: Do you not need a space here between image_tag and "loading_gif"?

Anbranin:
I'm gonna do this over.

neal-merkl:
Best of luck

PR #413, opened 2019-01-31
Ruby update

Anbranin:
Newer version of Ruby. This is good for now, more updates to follow.


Anbranin:
Forgot to add some things.

PR #418, opened 2019-02-26
Updates

Anbranin:
I updated rails. Enjoy.


app/models/hold.rb
```diff
@@ -20,9 +20,10 @@ class Hold < ActiveRecord::Base
   def handle_conflicting_rentals
     # if rental falls between data range of our hold
     # only reserved but not picked up
-    conflicting_rentals = Rental.reserved.joins('INNER JOIN rentals_items ON rentals_items.rental_id = rentals.id')
+    conflicting_rentals = Rental.reserved.joins(:rentals_items)
                                 .where('start_time <= :hold_end_time AND end_time >= :hold_start_time AND rentals_items.item_id = :hold_item_id',
                                        hold_start_time: start_time, hold_end_time: end_time, hold_item_id: item.id)
+    # Rental.where('start_time <= ? AND end_time >= ?', end_time, start_time) also works
```
nawilliams: Uneeded comment?

app/models/rental.rb
```diff
@@ -39,7 +40,7 @@ class Rental < ActiveRecord::Base
   def renter_is_assignable
     return unless renter && creator # will be validated elsewhere, dont add an error twice
     if creator.assignable_renters.exclude? renter
-      errors.add :renter, 'must have permission to assign'
+      errors.add :creator, 'must have permission to assign'
```
nawilliams: Shouldn't this still be renter? The renter is the one that it looks like the error is activating on.

app/models/rentals_item.rb
```diff
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
+
 class RentalsItem < ActiveRecord::Base
   has_paper_trail
   # use inverse of so we can create this with rental blank at first
   belongs_to :rental, inverse_of: :rentals_items
-  belongs_to :item
+  belongs_to :item, optional: true
+  # TODO: remove item_type from rentals_item.
```
nawilliams: Checking on this TODO

app/models/rentals_item.rb
```diff
@@ -1,11 +1,13 @@
...
   belongs_to :item_type
-  validates :rental, :item_type, presence: true # we dont validate item, that will be assigned by the rental
+  # we dont validate item, that will be assigned by the rental
```
nawilliams: Don't need comment if we removed item_type validation

nawilliams: Also even if we remove item_type, shouldn't rental still be validated?

bin/setup
```diff
@@ -1,29 +1,36 @@
 #!/usr/bin/env ruby
-require 'pathname'
+require 'fileutils'
+include FileUtils
```
nawilliams: Need both of these? Or is this a 5.2 thing?

bin/yarn
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
```
nawilliams: Rails includes yarn now?

config/environments/development.rb
```diff
@@ -9,47 +9,56 @@
...
 
-  # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
-
-  # Do care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.perform_deliveries = true
-  config.action_mailer.delivery_method = :smtp
-  config.action_mailer.smtp_settings = {
```
nawilliams: Removed smtp settings, not sure if needed.

config/environments/development.rb
```diff
@@ -9,47 +9,56 @@
...
   config.assets.debug = true
 
-  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
-  # yet still be able to expire them through the digest params.
-  config.assets.digest = true
-
-  # Adds additional error checking when serving assets at runtime.
-  # Checks for improperly declared sprockets dependencies.
-  # Raises helpful error messages.
-  config.assets.raise_runtime_errors = true
```
nawilliams: Baked into 5.2 or did we want to remove this?

config/environments/production.rb
```diff
@@ -31,16 +29,23 @@
...
 
+  # Enable serving of images, stylesheets, and JavaScripts from an asset server.
+  # config.action_controller.asset_host = 'http://assets.example.com'
+
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
+  # Store uploaded files on the local file system (see config/storage.yml for options)
+  config.active_storage.service = :local
```
nawilliams: Do we even have files that get uploaded to this app?

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
```
nawilliams: Unused?

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
+
```
nawilliams: Unused?

spec/factories/incurred_incidental.rb
```diff
@@ -20,7 +21,7 @@
   end
 
   factory :invalid_incidental, parent: :incurred_incidental do
-    rental_id nil
-    financial_transaction_attributes amount: 5
+    rental_id { nil }
+    # financial_transaction_attributes amount: 5
```
nawilliams: Delete if we don't want this.

spec/factories/rentals_item.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 FactoryBot.define do
   factory :rentals_item do
-    # no association to :rental, that would create an infinate loop
-    association :item
-    association :item_type
-    reservation_id { SecureRandom.uuid }
+    # no association to :rental, that would create an infinite loop
+    item_type
```
nawilliams: Not sure what this line does?

nawilliams (dismissed):
Looks good, only really pointed out things that looked a bit off that you might want to double check and a few leftover comments.

app/models/rental.rb
Anbranin: I actually interpreted this to mean "check that the person renting out this golf cart has permission to rent out this golf cart to this person" in which case 'renter' would not make sense here. Want me to go into detail on how I interpreted it?

app/models/rentals_item.rb
Anbranin: That's definitely not a TODO for this PR. It'll be later.

app/models/rentals_item.rb
Anbranin: All I did was not remove a comment. I don't know what they meant but I decided to tackle whatever they were talking about later, in which case the comment will be helpful.

bin/setup
Anbranin: Actually, `require` and `include` do two totally separate but rather similar things, so yes you do need both of them.

Anbranin: And it is just a 5.2 thing, this is a line added by Rails.

config/environments/production.rb
Anbranin: Rails 5.2 adds this line.

config/initializers/application_controller_renderer.rb
Anbranin: 5.2 added

config/initializers/content_security_policy.rb
Anbranin: Rails added this file too.

I should mention, there's an "rake rails update" task that adds these files. If you made a new 5.2 project it would include them.

spec/factories/rentals_item.rb
Anbranin: Same as  association :item_type.

config/environments/development.rb
Anbranin: Based on http://railsdiff.org/4.2.11/5.2.2 I think we're good

app/models/rental.rb
nawilliams: My only concern is the usefulness of the error. If we get this error, will it be easier to debug when it's added to the renter of the creator?

app/models/rentals_item.rb
nawilliams: Still leaves the question of removing the validation for rental, just want to double check that it was intended, as I only know you were removing item_type validations.

app/models/rental.rb
Anbranin: Well, since the method checks the permissions of the creator and the error mentions someone having permissions it really makes absolutely zero sense to have the renter's name here. The creator is assigning the rental to the renter, so "renter must have permission to assign" would not in this case be helpful.
The error message could be better, of course, because this has zero user-friendliness. If Jon King sees this message, he's going to come to IT. We'll know what's happening, but no user would.
Now, an error message like "sorry, you only have permission to assign rentals to users in your department" would be much clearer, but that isn't the point of this PR.

app/models/rentals_item.rb
Anbranin: `belongs_to` validates presence by default. So, I wasn't removing anything but redundancy.

nawilliams (dismissed):
Comments addressed. Looks good.

elisacoutant (approved):

